PR #24, opened 2020-11-28
Hotfix/8 nested input filter config

alex-patterson-webdev:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | maybe
| New Feature   | no
| RFC           | no

### Description

This pull request aids to address the problem defined in issue #8 

I have added the ability to nest the input specification of nested input filters, when calling `Factory::createInputFilter()` 

Currently, when nesting input filter configuration, the `type` key sits at the same level as the names of the required inputs. This can cause issue if the input you wish to add also has the name `type`,

    $spec = [
        'baz' => [
            'type' => InputFilter::class,
            'foo' => [
                'name' => 'foo',
                'required' => false,
                'validators' => [
                    [
                       'name' => Validator\NotEmpty::class,
                    ],
                ],
            ],
        ],
    ];

    $factory->createInputFilter($spec);

This PR allows the optional `inputs` key that nests the relevant inputs. If we have an input with the name `type`  if will no longer conflict.

    $spec = [
        'baz' => [
            'type' => InputFilter::class,
            'inputs' => [
                'type' => [
                    'name' => 'foo',
                    'required' => false,
                    'validators' => [
                        [
                            'name' => Validator\NotEmpty::class,
                        ],
                    ],
                ],
            ],
        ],
    ];

    $factory->createInputFilter($spec);

# BC Consideration

There is a possibility that existing code has an input already named `inputs` so we could consider changing this to something more unique

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -324,6 +324,10 @@ class Factory
             return $inputFilter;
         }
 
+        if (isset($inputFilterSpecification['inputs']) && is_array($inputFilterSpecification['inputs'])) {
+            $inputFilterSpecification = $inputFilterSpecification['inputs'];
+        }
+
         foreach ($inputFilterSpecification as $key => $value) {
             if (null === $value) {
                 continue;
--- a/test/FactoryTest.php
+++ b/test/FactoryTest.php
@@ -503,82 +503,15 @@ class FactoryTest extends TestCase
         $this->assertEquals($input, $inputFilter->get('foo'));
     }
 
-    public function testFactoryWillCreateInputFilterAndAllInputObjectsFromGivenConfiguration()
+    /**
+     * @param array $spec
+     *
+     * @dataProvider getFactoryWillCreateInputFilterAndAllInputObjectsFromGivenConfigurationData
+     */
+    public function testFactoryWillCreateInputFilterAndAllInputObjectsFromGivenConfiguration(array $spec)
     {
         $factory     = $this->createDefaultFactory();
-        $inputFilter = $factory->createInputFilter([
-            'foo' => [
-                'name'       => 'foo',
-                'required'   => false,
-                'validators' => [
-                    [
-                        'name' => Validator\NotEmpty::class,
-                    ],
-                    [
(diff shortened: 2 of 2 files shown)


froschdesign:
@alex-patterson-webdev 
> This can cause issue if the input you wish to add also has the name `type`

Maybe I have missed something, but you do not need to specify a key:


```php
$spec = [
    'group' => [
        'type' => Laminas\InputFilter\InputFilter::class,
        [
            'name'       => 'type',
            'required'   => false,
            'validators' => [
                [
                    'name' => Laminas\Validator\NotEmpty::class,
                ],
            ],
        ],
    ],
];

$inputFilter = (new Laminas\InputFilter\Factory())->createInputFilter($spec);

var_dump($inputFilter->get('group')->get('type')->getName()); // 'type'
```

In this case the input filter uses the value of the name property of the input:

https://github.com/laminas/laminas-inputfilter/blob/22ae90c7178cac7ddad803b73cce50bea9eff697/src/BaseInputFilter.php#L95-L97

froschdesign:
I personally find the naming in this example very confusing:

> ```php
> $spec = [
>     'baz' => [
>         'type' => InputFilter::class,
>         'inputs' => [
>             'type' => [
>                 'name' => 'foo',
>                 'required' => false,
>                 'validators' => [
>                     [
>                         'name' => Validator\NotEmpty::class,
>                     ],
>                 ],
>             ],
>         ],
>     ],
> ];
> ```

Because this means:

```php
var_dump($inputFilter->get('baz')->get('type')->getName()); // foo
```

alex-patterson-webdev:
@froschdesign Thanks, I came across this issue today and google took me to issue #8. I had just assumed it wasn't possible.

It seems you are correct that the inputs can be specified with numeric keys; providing they have a `name`. The section of code is in fact [near where I added my changes](https://github.com/laminas/laminas-inputfilter/blob/2.11.x/src/Factory.php#L347) so I perhaps should have noticed. Anyway, this is great and a better solution. Thanks for your time.

PR #40, opened 2021-10-29
ArrayAccess arg type in BaseInputFilter::validateInputs()

darckking:
Signed-off-by: illia <illia.gapak@ffflabel.com>

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

I described the problem on Laminas forums, you can find it here
https://discourse.laminas.dev/t/arrayaccess-arg-type-in-baseinputfilter-validateinputs/2567
The test I added illustrates 2 problems:
1. Wrong behavior on PHP >7.4 because array_key_exists doesn't call for ArrayAccess::offsetExists()
2. Deprecation warning on PHP 7.4, I believe PHP 8 will throw an error, haven't tested that.

At least I expect that ArrayAccess will be removed from DocBlock of BaseInputFilter::validateInputs(), but ideally validateInputs() $data argument must be typed as an array. When min PHP version will change to 7.4 - BaseInputFilter::$data typed as an array as well. After this fixed - my test has to be removed or BaseInputFilter::validateInputs() must get proper support for ArrayAccess object.

### DIFF
--- a/test/BaseInputFilterTest.php
+++ b/test/BaseInputFilterTest.php
@@ -683,6 +683,52 @@ class BaseInputFilterTest extends TestCase
         self::assertSame($filteredArray, $baseInputFilter->getRawValues());
     }
 
+    public function testValidateInputsUsingArrayAccessObjectAsDataArgument()
+    {
+        $myInputFilter = new class extends BaseInputFilter{
+            protected function validateInputs(array $inputs, $data = [], $context = null)
+            {
+                // according to DocBlock $data must be array or ArrayAccess, so if it's not an ArrayAccess then array
+                // and I can safely pass $data to MyArrayAccess object __construct
+                if (! $data instanceof \ArrayAccess) {
+                    $data = new class($data) implements \ArrayAccess {
+                        protected $input1 = 'default';
+                        public function __construct(array $data)
+                        {
+                            if (key_exists('input1', $data)) {
+                                $this->input1 = $data['input1'];
+                            }
+                        }
+                        public function offsetExists($offset)
+                        {
+                            return $offset === 'input1';
+                        }
+                        public function offsetGet($offset)
+                        {
+                            return $this->$offset;
+                        }
+                        public function offsetSet($offset, $value)
+                        {
+                        }
+                        public function offsetUnset($offset)
+                        {
+                        }
+                    };
+                }
+                // So I pass ArrayAccess object to parent function, what is allowed according to DocBlock
+                return parent::validateInputs($inputs, $data, $context);
+            }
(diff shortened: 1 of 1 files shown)


froschdesign:
@darckking 
Please check the failed checks:

* https://github.com/laminas/laminas-inputfilter/runs/4048519547?check_suite_focus=true#step:3:309
* https://github.com/laminas/laminas-inputfilter/runs/4048519683?check_suite_focus=true#step:3:350
* https://github.com/laminas/laminas-inputfilter/runs/4048519768?check_suite_focus=true#step:3:350

Thanks in advance! 👍

darckking:
> @darckking Please check the failed checks:
> 
> * https://github.com/laminas/laminas-inputfilter/runs/4048519547?check_suite_focus=true#step:3:309
> * https://github.com/laminas/laminas-inputfilter/runs/4048519683?check_suite_focus=true#step:3:350
> * https://github.com/laminas/laminas-inputfilter/runs/4048519768?check_suite_focus=true#step:3:350
> 
> Thanks in advance! 👍

darckking:
@froschdesign 
Hi
What exactly should I check ?
Yes, the test I added fails. But that is my goal to show that there is a problem. When I use ArrayAccess argument which is allowed according to annotation.
7.3 - wrong result.
7.4  - deprecation warning
8 - type error

froschdesign:
@darckking 
> Yes, the test I added fails. But that is my goal to show that there is a problem.

You are right. Thank you for the reminder!
Would you also provide a solution?

darckking:
@froschdesign 

> At least I expect that ArrayAccess will be removed from DocBlock of BaseInputFilter::validateInputs(), but ideally validateInputs() $data argument must be typed as an array. When min PHP version will change to 7.4 - BaseInputFilter::$data typed as an array as well. After this fixed - my test has to be removed or BaseInputFilter::validateInputs() must get proper support for ArrayAccess object.

Basically it depends on how important it is to have ArrayAccess as argument. If no - then remove it, if yes - then ArrayAccess::offsetExists() must be used instead of array_key_exists() in BaseInputFilter line 261

froschdesign:
@darckking 
> …then remove it…

It is a bug in DocBlocks, so go for it.
Thanks in advance! 👍

darckking:
I removed ArrayAccess and created another pull request [#46 ](https://github.com/laminas/laminas-inputfilter/pull/46).
I close this pull request because the test I added is not relevant anymore after ArrayAccess removal.

PR #43, opened 2021-11-09
CollectionInputFilter::isValid($context) - pass $context to nested input filter

### LINKED ISSUE #42: CollectionInputFilter - isValid does not pass $context variable
[CollectionInputFilter::isValid($context = null)](https://github.com/laminas/laminas-inputfilter/blob/2.13.x/src/CollectionInputFilter.php#L211) does not pass $context variable to [nested input filter](https://github.com/laminas/laminas-inputfilter/blob/2.13.x/src/CollectionInputFilter.php#L240). I do not know if it is expected behavior.

I need to get whole input filter $context from my nested validators. If it is bug, I can create PR. If it is expected behavior I can create PR with new ContextCollectionInputFilter which will pass $context to nested input filter.

vaclavvanik:
As I mentioned in #42 `CollectionInputFilter::isValid($context)` does not pass given `$context` to nested input filter.

I expect the current behavior is rather bug - php doc - @param mixed $context Ignored, but present to retain signature compatibility.

This PR fixes passing `$context` and thus `$context` is available in nested input filter and its validators.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -921,7 +921,7 @@
       <code>function () use ($dataRaw, $dataFiltered) {</code>
       <code>function () use ($dataRaw, $dataFiltered, $errorMessage) {</code>
     </MissingClosureReturnType>
-    <MissingParamType occurrences="21">
+    <MissingParamType occurrences="19">
       <code>$count</code>
       <code>$count</code>
       <code>$count</code>
@@ -941,10 +941,8 @@
       <code>$inputFilter</code>
       <code>$required</code>
       <code>$value</code>
-      <code>$customContext</code>
-      <code>$expectedContext</code>
     </MissingParamType>
-    <MissingReturnType occurrences="31">
+    <MissingReturnType occurrences="29">
       <code>countVsDataProvider</code>
       <code>dataNestingCollection</code>
       <code>dataVsValidProvider</code>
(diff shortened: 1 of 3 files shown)


src/CollectionInputFilter.php
```diff
@@ -206,7 +206,6 @@ public function setNotEmptyValidator(NotEmpty $notEmptyValidator)
 
     /**
      * {@inheritdoc}
-     * @param mixed $context Ignored, but present to retain signature compatibility.
```
froschdesign, line 209: This part should be updated and not removed.

vaclavvanik: Imho this line should be removed because is same as in [BaseInputFilter](https://github.com/laminas/laminas-inputfilter/blob/2.13.x/src/BaseInputFilter.php#L218)

froschdesign: Then the rest of DocBlock must be updated:

```php
/**
 * @inheritdoc
 */
```

psalm-baseline.xml
```diff
@@ -972,6 +974,8 @@
       <code>testSettingDataAsTraversableWithInvalidCollectionsRaisesException</code>
       <code>testSettingDataWithNonArrayNonTraversableRaisesException</code>
       <code>testUsesMessageFromComposedNotEmptyValidatorWhenRequiredButCollectionIsEmpty</code>
+      <code>contextProvider</code>
+      <code>testValidationContext</code>
```
froschdesign: Uff, Psalm was added but without any other changes.
For new code, it should be done right away. Please add the missing return and parameter types in the code itself.

PR #46, opened 2021-11-28
Bugfix: removed ArrayAccess from BaseInputFilter::validateInputs() DocBlocks

darckking:
Signed-off-by: illia <illia.gapak@ffflabel.com>

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

I described the problem on [Laminas Forums](https://discourse.laminas.dev/t/arrayaccess-arg-type-in-baseinputfilter-validateinputs/2567) and created a test to show the issue [#40 ](https://github.com/laminas/laminas-inputfilter/pull/40)

This is breaking change for everyone who extends BaseInputFilter::validateInputs() due function signature change - I added 'array' type to $data argument.

### DIFF
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -229,9 +229,9 @@ class BaseInputFilter implements
      * @param  mixed|null $context
      * @return bool
      */
-    protected function validateInputs(array $inputs, array $data = [], $context = null)
+    protected function validateInputs(array $inputs, $data = [], $context = null)
     {
-        $inputContext = $context ?: (array_merge($this->getRawValues(), $data));
+        $inputContext = $context ?: (array_merge($this->getRawValues(), (array) $data));
 
         $this->validInputs   = [];
         $this->invalidInputs = [];
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -2,7 +2,6 @@
 
 namespace Laminas\InputFilter;
 
-use ArrayAccess;
 use Laminas\Stdlib\ArrayUtils;
 use Laminas\Stdlib\InitializableInterface;
 use Traversable;
@@ -226,13 +225,13 @@ class BaseInputFilter implements
      * Validate a set of inputs against the current data
      *
      * @param  string[] $inputs Array of input names.
-     * @param  array|ArrayAccess $data
+     * @param  array $data
      * @param  mixed|null $context
      * @return bool
      */
-    protected function validateInputs(array $inputs, $data = [], $context = null)
+    protected function validateInputs(array $inputs, array $data = [], $context = null)
     {
-        $inputContext = $context ?: (array_merge($this->getRawValues(), (array) $data));
+        $inputContext = $context ?: (array_merge($this->getRawValues(), $data));
 
         $this->validInputs   = [];
         $this->invalidInputs = [];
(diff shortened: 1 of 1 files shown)


src/BaseInputFilter.php
```diff
@@ -226,13 +225,13 @@ public function isValid($context = null)
      * Validate a set of inputs against the current data
      *
      * @param  string[] $inputs Array of input names.
-     * @param  array|ArrayAccess $data
+     * @param  array $data
      * @param  mixed|null $context
      * @return bool
      */
-    protected function validateInputs(array $inputs, $data = [], $context = null)
+    protected function validateInputs(array $inputs, array $data = [], $context = null)
```
Ocramius, line 232: This is a clear BC break, as it forces callers of this method to change.

The old signature needs to be kept, with an `instanceof` check.

Ocramius (changes requested):

src/BaseInputFilter.php
```diff
@@ -226,7 +225,7 @@ public function isValid($context = null)
      * Validate a set of inputs against the current data
      *
      * @param  string[] $inputs Array of input names.
-     * @param  array|ArrayAccess $data
```
Ocramius, line 229: As it stands, this is still a BC break due to change of declaration

Ocramius (changes requested):

src/BaseInputFilter.php
darckking: I'm a bit confused. @froschdesign told to me that I can remove ArrayAccess, [here](https://github.com/laminas/laminas-inputfilter/pull/40#issuecomment-981098629)

Ocramius: Eh, it's still `protected` API (which counts as `public`, because of inheritance being part of the design of  `BaseInputFilter`.

Perhaps there was some misunderstanding?

froschdesign: @Ocramius 
The DocBlock is wrong here because an ArrayAccess object has never worked with this method.

darckking: Yeah
this is what I wrote 

> Basically it depends on how important it is to have ArrayAccess as argument. If no - then remove it, if yes - then ArrayAccess::offsetExists() must be used instead of array_key_exists() in BaseInputFilter line 261

Then it should be like you say, keep ArrayAccess, and add instanceof check. Right ?

Ocramius: Makes sense, thanks - that's misunderstanding on my end then :)

Ocramius (approved):

src/BaseInputFilter.php
Ocramius: Nope, all good - I was wrong here, 🚢

Ocramius:
We can remove the `array` cast too: 

```


░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░
Error: Redundant cast to array given docblock-provided type
```

darckking:
Yes
Maybe I can revert the last commit or 'array' type for $data arg is still a problem ? Because it is a BC break, but feels like a right thing to do.

Ocramius:
> Because it is a BC break, but feels like a right thing to do.

Yeah, seems like the right move - let's do it.

Note: child classes overriding this would still work as expected, because as of older PHP 7 versions, parameter variance can change with laxer types (or no type)

PR #48, opened 2021-12-02
Provide PHP 8.1 support

### LINKED ISSUE #44: PHP 8.1 support
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes

#### Summary

PHP with [version 8.1 is released](https://www.php.net/releases/8.1/index.php) and for this component the support of the new version is still missing.

weierophinney:
This patch provides PHP 8.1 support by making the following changes:

- Adds 8.1.0 to list of PHP constraints
- Bumps laminas-db to 2.13.4
- Bumps laminas-filter to 2.13
- Bumps laminas-validator to 2.15
- Bumps PHPUnit to 9.5.5+
- Bumps Prophecy to 1.14+
- Switches from laminas-zendframework-bridge + replace rule to conflict rule
- Ignores UnusedClass errors that reference ZF classes
- Adds `#[ReturnTypeWillChange]` attribute to `BaseInputFilter::count()` implementation
- Ignores platform requirements when testing for PHP 8.1 (primarily due to laminas-coding-standard version)

### DIFF
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -1,0 +1,5 @@
+{
+    "ignore_php_platform_requirements": {
+        "8.1": true
+    }
+}
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,13 +1,19 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd" bootstrap="vendor/autoload.php" colors="true">
-  <coverage processUncoveredFiles="true">
-    <include>
-      <directory suffix=".php">src</directory>
-    </include>
-  </coverage>
-  <testsuites>
-    <testsuite name="laminas-inputfilter Test Suite">
-      <directory>./test</directory>
-    </testsuite>
-  </testsuites>
+<phpunit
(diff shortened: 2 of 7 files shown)


.laminas-ci.json
```diff
@@ -0,0 +1,5 @@
+{
+    "ignore_php_platform_requirements": {
```
weierophinney, line 2: This is due to being on the laminas-coding-standard 2.2 series. I didn't bump to 2.3, because that version adds the `declare(strict_types=1)` directive and strict equality checks by default, which can have a fair number of knock-on effects in a library like this.

froschdesign (approved):

phpunit.xml.dist
```diff
@@ -1,13 +1,19 @@
...
-  <testsuites>
-    <testsuite name="laminas-inputfilter Test Suite">
-      <directory>./test</directory>
-    </testsuite>
-  </testsuites>
+<phpunit
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
+    bootstrap="vendor/autoload.php"
+    convertDeprecationsToExceptions="true"
```
weierophinney, line 6: Added this, as PHPUnit 9.5.10 toggles this off by default, and we want to detect deprecations early.

froschdesign: Reference: https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-9.5.md#9510---2021-09-25

PR #49, opened 2022-01-05
Introduce allowNull setting for Input.

darckking:
Signed-off-by: illia <illia.hapak@scharc.com>

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

I want propose alternative for deprecated allowEmpty and continueIfEmpty settings.
Why ?
Let's think about following example. We have 3 entities: Store, Product, Category. Product has Category as optional relation. Categories have relation to Store.
```php
$inputFilterSpecs = [
    //... the other inputs like name, cost, etc
    'category' => [
        'name' => 'category',
        'required' => false,
        'filters' => [
            ['name' => EntityFilter::class], // find Category entity in DB by id
        ],
        'validators' => [
            [
                'name' => ObjectExistsValidator::class, // make sure that Category exists before continue
                'break_chain_on_failure' => true,
            ],
            ['name' => CategoryValidator::class], // check if user can add product to this category, etc
        ],
    ],
];
```
I set data to Product using Laminas\Hydrator\ClassMethodsHydrator
```php
class Product {
    ...
    public function setCategory(?Category $category){
        $this->category = $category
    }
    ...
}
```
Problem of 'category' input: 
required = false AND allowEmpty = false AND continueIfEmpty = false. 
As a consequence empty array "[]" and empty string "" are accepted as valid values and lead to **TypeError**.

To avoid this I can add 'continue_if_empty' => true XOR make input required and force client explicitly send null value. But it means that null will be filtered and validated.

Solution using $allowNull setting
```php
$inputFilterSpecs = [
    //... the other inputs like name, cost, etc
    'category' => [
        'name' => 'category',
        'required' => false,
        'allow_null' => true,
        'continue_if_empty' => true,
        // ... filter, validators, etc
    ],
];
```

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -254,6 +254,9 @@ class Factory
                     }
                     $this->populateValidators($input->getValidatorChain(), $value);
                     break;
+                case 'allow_null':
+                    $input->setAllowNull($value);
+                    break;
                 default:
                     // ignore unknown keys
                     break;
--- a/src/Input.php
+++ b/src/Input.php
@@ -28,6 +28,8 @@ class Input implements
      */
     protected $continueIfEmpty = false;
 
+    protected bool $allowNull = false;
+
     /** @var bool */
     protected $breakOnFailure = false;
 
@@ -106,6 +108,12 @@ class Input implements
     public function setContinueIfEmpty($continueIfEmpty)
     {
         $this->continueIfEmpty = (bool) $continueIfEmpty;
+        return $this;
+    }
+
+    public function setAllowNull(bool $allowNull): InputInterface
+    {
+        $this->allowNull = $allowNull;
         return $this;
     }
 
@@ -230,6 +238,11 @@ class Input implements
         return $this->continueIfEmpty;
     }
 
+    public function allowNull(): bool
+    {
+        return $this->allowNull;
+    }
+
     /**
      * @return string|null
      */
@@ -289,6 +302,9 @@ class Input implements
      */
     public function getValue()
     {
+        if($this->allowNull && is_null($this->value)){
+            return null;
+        }
         $filter = $this->getFilterChain();
         return $filter->filter($this->value);
     }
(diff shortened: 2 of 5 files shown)


froschdesign:
I think your problem can be solved with other options like the [`ToNull` filter](https://docs.laminas.dev/laminas-filter/standard-filters/#tonull) or a fallback value, no need to extend the input filter.

darckking:
ToNull filter -  Let's assume I put ToNull filter before EntityFilter in my example.
```php
$inputFilterSpecs = [
    //... the other inputs like name, cost, etc
    'category' => [
        'name' => 'category',
        'required' => false,
        'filters' => [
            ['name' => ToNull::class],
            ['name' => EntityFilter::class], // find Category entity in DB by id
        ],
    ],
];
```
Indeed, it will filter [] or "" to null and will solve TypeError. However it works only for not required properties. If we make property required then ValidatorChain will be executed, to bypass it - allowEmpty (deprecated) must be set to true. Or validators must be configurable to consider null as valid value - I don't like this option, such configuration must be on Input not validators.

Another problem with this is that API is not strict enough, when clients post [] or "" they should get validation error, that only integers or null allowed. Also Imo it's better to skip execution of FilterChain and ValidatorChain when we know that null is a valid value and no need to filter / validate it, in same manner as continueIfEmpty = false; does.

Fallback value is even worse solution, when input has fallback value and ValidatorChain is invalid, it silently sets fallback value as input's value. So API clients think they can post anything and API will swallow it.

So yeah, these are workarounds but not acceptable solutions for me.

froschdesign:
> …these are workarounds but not acceptable solutions for me.

An the extension of the input filter class is also a workaround and not the desired way.

vaclavvanik:
Use required true. Set NotEmptyValidator with settings that allow null. (I am answering from phone thusI cannot paste example)

darckking:
Good idea. Will try that. 
Thank you!

PR #53, opened 2022-06-07
Remove ServiceManager V2 compatibility

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| QA            | yes

### Description

- Removes SMv2 compat in the InputFilterPluginManager because it cannot possibly be installed anymore
- Replaces usage of aliases for Input|Filter|Validator plugin managers with FQCN
- Adds a psalm plugin to dev dependencies for typical Psr Container methods
- Updates locked dependencies on 7.4
- Fixes a number of psalm issues, updating baseline


docs/book/specs.md
```diff
@@ -74,7 +75,7 @@ class MyValidatingControllerFactory implements FactoryInterface
         $services = $controllers->getServiceLocator();
 
         // Retrieve the InputFilterManager
-        $filters = $services->get('InputFilterManager');
+        $filters = $services->get(InputFilterPluginManager::class);
```
gsteel, line 78: Update the docs so that they prefer fetching the plugin manager by FQCN

psalm.xml.dist
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <psalm
+    errorLevel="1"
```
gsteel, line 3: Fix deprecated option since a few versions ago

src/InputFilterAbstractServiceFactory.php
```diff
@@ -60,11 +59,6 @@ public function canCreate(ContainerInterface $services, $rName)
      */
     public function canCreateServiceWithName(ServiceLocatorInterface $container, $name, $requestedName)
```
gsteel, line 63: Should I possibly mark these legacy methods as deprecated also?

composer.json
```diff
@@ -40,6 +40,7 @@
     "require-dev": {
         "laminas/laminas-coding-standard": "~2.2.1",
         "laminas/laminas-db": "^2.13.4",
+        "lctrs/psalm-psr-container-plugin": "^1.6",
```
Ocramius: I don't think we should add this for the sake of "winning against static analysis holes" :thinking:

src/InputFilterPluginManager.php
```diff
@@ -71,17 +77,9 @@ class InputFilterPluginManager extends AbstractPluginManager
     protected $sharedByDefault = false;
 
     /**
-     * Whether or not to share by default (v2)
-     *
-     * @var bool
-     */
-    protected $shareByDefault = false;
```
gsteel, line 78: Perhaps dropping this is a BC break. Should I revert and mark the class as @\final?

composer.json
Ocramius: A psalm `.stub.php` file would suffice for our use-case, perhaps?

src/InputFilterPluginManagerFactory.php
```diff
@@ -2,28 +2,39 @@
...
 use Laminas\ServiceManager\FactoryInterface;
 use Laminas\ServiceManager\ServiceLocatorInterface;
+use Laminas\ServiceManager\ServiceManager;
+use Psr\Container\ContainerInterface as PsrContainer;
 
+use function assert;
 use function is_array;
 
+/**
+ * @link ServiceManager
```
gsteel, line 17: Prevents automated removal of unused use by phpcbf

src/InputFilterPluginManager.php
```diff
@@ -71,17 +77,9 @@ class InputFilterPluginManager extends AbstractPluginManager
...
-
-    /**
-     * @param null|ConfigInterface|ContainerInterface $configOrContainer For laminas-servicemanager v2,
-     *     null or a ConfigInterface instance are allowed; for v3, a
-     *     ContainerInterface is expected.
-     * @param array $v3config Optional configuration array (laminas-servicemanager v3 only)
+     * @param null|ConfigInterface|ContainerInterface|PsrContainerInterface $configOrContainer
+     * @param array $v3config
```
Ocramius: Can be removed

test/InputFilterPluginManagerCompatibilityTest.php
```diff
@@ -31,4 +31,10 @@ protected function getInstanceOf()
     {
         // InputFilterManager accepts multiple instance types
     }
+
+    public function testShareByDefaultAndSharedByDefault(): void
+    {
+        // The plugin manager no longer supports V2 so this inherited test is forced to pass
+        self::assertTrue(true);
```
Ocramius: This shouldn't be added as-is :D

Ocramius (changes requested):

composer.json
gsteel: > I don't think we should add this for the sake of "winning against static analysis holes" 🤔

Is that because of the dependency burden? I frequently use this plugin in projects because it avoids assigning the service to a temporary variable along with an `assert instanceof` - Not familiar with `.stub.php` - I'll have a Google…

Ocramius: Yes, adding dependencies should be done with a lot of care, since most of our work right now is playing with the dependency puzzle until it clicks together.

Having a bigger baseline is still better than having more dependencies

gsteel: @Ocramius - trying to get the stub right but can't figure out how to get psalm to return instance of the given class-string: https://psalm.dev/r/93dc7fec74 - Any chance you could help me out there?

Ocramius: See https://psalm.dev/r/027e9ad09e

The `T` needs to be on the method in this case, since it's not the whole container that is templated with `T`

gsteel: 🙏

Ocramius: Just cross-linking: https://github.com/Lctrs/psalm-psr-container-plugin/issues/545

.stub.php
```diff
@@ -0,0 +1,19 @@
+<?php
```
Ocramius: I think we should document this stub a bit:

 * file name more descriptive
 * a few comments here, explaining our intent

.stub.php
```diff
@@ -0,0 +1,19 @@
...
+
+declare(strict_types=1);
+
+namespace Psr\Container {
+
+    interface ContainerInterface
+    {
+        /** @param string|class-string $name */
```
Ocramius: In theory, we can drop this declaration completely, since the original class declaration should take over :thinking: 

Can you give it a try? Let's keep all of this very minimal

src/InputFilterAbstractServiceFactory.php
Ocramius: Deprecation will certainly allow for more graceful removal, yes.

test/InputTest.php
```diff
@@ -48,14 +48,26 @@ public function assertRequiredValidationErrorMessage(Input $input, string $messa
         $message  = $message ?: 'Expected failure message for required input';
         $message .= ';';
 
-        $messages = $input->getMessages();
-        $this->assertIsArray($messages, $message . ' non-array messages array');
+        $expectedKey = NotEmptyValidator::IS_EMPTY;
+        $messages    = $input->getMessages();
+        /** @psalm-suppress RedundantConditionGivenDocblockType */
```
Ocramius: We can drop this assertion overall, no?

Ocramius (changes requested):

.stub.php
gsteel: It doesn't appear the declaration can be dropped - looks like the stub replaces the interface entirely.

Ocramius (approved):
Thanks @gsteel!

PR #54, opened 2022-06-09
Static analysis Improvements and general clean up

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | Hopefully Not
| QA            | yes

### Description

Sorry for the size of this patch - My original intention was to improve type inference for some seemingly simple stuff but things quickly got out of hand.

I'll do a review and point out the stuff I think might be risky.

Hopefully this will provide decent type checking for input filter specification arrays along with some other improvements

### DIFF
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -180,7 +180,7 @@ class BaseInputFilter implements
     /**
      * Set data to use when validating and filtering
      *
-     * @param  iterable<array-key, mixed>|null $data null is cast to an empty array.
+     * @param  iterable|null $data null is cast to an empty array.
      * @throws Exception\InvalidArgumentException
      * @return InputFilterInterface
      */
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -183,7 +183,6 @@ class BaseInputFilter implements
      * @param  iterable<array-key, mixed>|null $data null is cast to an empty array.
      * @throws Exception\InvalidArgumentException
      * @return InputFilterInterface
-     * @psalm-assert array<array-key, mixed> $this->data
      */
     public function setData($data)
     {
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -13,6 +13,7 @@ use function array_intersect;
 use function array_key_exists;
 use function array_keys;
 use function array_merge;
+use function assert;
 use function count;
 use function func_get_args;
 use function get_class;
@@ -23,9 +24,6 @@ use function is_object;
 use function is_string;
 use function sprintf;
 
-/**
- * @psalm-import-type InputData from InputFilterInterface
- */
 class BaseInputFilter implements
     InputFilterInterface,
     UnknownInputsCapableInterface,
@@ -33,10 +31,10 @@ class BaseInputFilter implements
     ReplaceableInputInterface,
     UnfilteredDataInterface
 {
-    /** @var InputData|null */
+    /** @var array<array-key, mixed>|null */
     protected $data;
 
-    /** @var InputData */
+    /** @var array<array-key, mixed> */
     protected $unfilteredData = [];
 
     /** @var array<string, InputInterface|InputFilterInterface> */
@@ -184,16 +182,15 @@ class BaseInputFilter implements
     /**
      * Set data to use when validating and filtering
      *
-     * @param  InputData|Traversable<string, mixed>|null $data null is cast to an empty array.
+     * @param  array<array-key, mixed>|Traversable<array-key, mixed>|null $data null is cast to an empty array.
      * @throws Exception\InvalidArgumentException
      * @return InputFilterInterface
-     * @psalm-assert InputData $this->data
+     * @psalm-assert array<array-key, mixed> $this->data
      */
     public function setData($data)
     {
         // A null value indicates an empty set
         if (null === $data) {
-            /** @psalm-var InputData $data */
             $data = [];
         }
 
@@ -461,7 +458,7 @@ class BaseInputFilter implements
      * List should be an associative array of named input/value pairs,
      * with the values unfiltered.
      *
-     * @return InputData|array<string, InputData>
+     * @return array<array-key, mixed>
      */
     public function getRawValues()
     {
@@ -523,6 +520,7 @@ class BaseInputFilter implements
      */
     protected function populate()
     {
+        assert($this->data !== null);
         foreach (array_keys($this->inputs) as $name) {
             $input = $this->inputs[$name];
 
@@ -547,7 +545,7 @@ class BaseInputFilter implements
                 continue;
             }
 
-            /** @psalm-var array|mixed $value */
+            /** @psalm-var mixed $value */
             $value = $this->data[$name];
 
             if ($input instanceof InputFilterInterface) {
@@ -555,8 +553,8 @@ class BaseInputFilter implements
                 if (! is_array($value) && ! $value instanceof Traversable) {
                     $value = [];
                 }
+                /** @psalm-var array<array-key, mixed> $value */
 
-                /** @psalm-var InputData $value */
                 $input->setData($value);
                 continue;
             }
@@ -631,7 +629,7 @@ class BaseInputFilter implements
     }
 
     /**
-     * @return InputData
+     * @return array<array-key, mixed>
      */
     public function getUnfilteredData()
     {
@@ -639,7 +637,7 @@ class BaseInputFilter implements
     }
 
     /**
-     * @param InputData $data
+     * @param array<array-key, mixed> $data
      * @return $this
      */
     public function setUnfilteredData($data)
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -43,7 +43,7 @@ class BaseInputFilter implements
     /** @var array<string, InputInterface|InputFilterInterface>|null */
     protected $invalidInputs;
 
-    /** @var null|list<string> Input names */
+    /** @var null|array<array-key, string> Input names */
     protected $validationGroup;
 
     /** @var array<string, InputInterface|InputFilterInterface>|null */
@@ -106,8 +106,6 @@ class BaseInputFilter implements
             $original->merge($input);
             return $this;
         }
-
-        /** @psalm-var string $name */
 
         $this->inputs[$name] = $input;
         return $this;
@@ -238,7 +236,7 @@ class BaseInputFilter implements
     /**
      * Validate a set of inputs against the current data
      *
-     * @param  list<string> $inputs Array of input names.
+     * @param  array<array-key, string> $inputs Array of input names.
      * @param  array<array-key, mixed> $data
      * @param  mixed|null $context
      * @return bool
@@ -306,7 +304,7 @@ class BaseInputFilter implements
      * Implementations should allow passing a single array value, or multiple arguments,
      * each specifying a single input.
      *
-     * @param  string|list<string>|array<string, list<string>> $name
+     * @param  string|array<array-key, string> $name
      * @throws Exception\InvalidArgumentException
      * @return InputFilterInterface
      */
@@ -341,8 +339,6 @@ class BaseInputFilter implements
         } else {
             $inputs = func_get_args();
         }
-
-        /** @psalm-var list<string> $inputs */
 
         if (! empty($inputs)) {
             $this->validateValidationGroup($inputs);
@@ -418,7 +414,6 @@ class BaseInputFilter implements
         foreach ($inputs as $name) {
             $input = $this->inputs[$name];
 
-            /** @psalm-var mixed $value */
             $value = $input instanceof InputFilterInterface
                 ? $input->getValues()
                 : $input->getValue();
@@ -496,10 +491,9 @@ class BaseInputFilter implements
     /**
      * Ensure all names of a validation group exist as input in the filter
      *
-     * @param  array $inputs Input names
+     * @param  array<array-key, string> $inputs Input names
      * @return void
      * @throws Exception\InvalidArgumentException
-     * @psalm-assert list<string> $inputs
      */
     protected function validateValidationGroup(array $inputs)
     {
@@ -545,7 +539,7 @@ class BaseInputFilter implements
                 continue;
             }
 
-            /** @psalm-var mixed $value */
+            /** @psalm-suppress MixedAssignment */
             $value = $this->data[$name];
 
             if ($input instanceof InputFilterInterface) {
@@ -553,7 +547,6 @@ class BaseInputFilter implements
                 if (! is_array($value) && ! $value instanceof Traversable) {
                     $value = [];
                 }
-                /** @psalm-var iterable<array-key, mixed> $value */
 
                 $input->setData($value);
                 continue;
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -182,7 +182,7 @@ class BaseInputFilter implements
     /**
      * Set data to use when validating and filtering
      *
-     * @param  array<array-key, mixed>|Traversable<array-key, mixed>|null $data null is cast to an empty array.
+     * @param  iterable<array-key, mixed>|null $data null is cast to an empty array.
      * @throws Exception\InvalidArgumentException
      * @return InputFilterInterface
      * @psalm-assert array<array-key, mixed> $this->data
@@ -553,7 +553,7 @@ class BaseInputFilter implements
                 if (! is_array($value) && ! $value instanceof Traversable) {
                     $value = [];
                 }
-                /** @psalm-var array<array-key, mixed> $value */
+                /** @psalm-var iterable<array-key, mixed> $value */
 
                 $input->setData($value);
                 continue;
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -20,8 +20,12 @@ use function gettype;
 use function is_array;
 use function is_int;
 use function is_object;
+use function is_string;
 use function sprintf;
 
+/**
+ * @psalm-import-type InputData from InputFilterInterface
+ */
 class BaseInputFilter implements
     InputFilterInterface,
     UnknownInputsCapableInterface,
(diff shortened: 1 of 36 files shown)


src/BaseInputFilter.php
```diff
@@ -506,6 +520,7 @@ protected function validateValidationGroup(array $inputs)
      */
     protected function populate()
     {
+        assert($this->data !== null);
```
gsteel, line 516: This method is only ever called once data is non-null

src/CollectionInputFilter.php
```diff
@@ -20,34 +21,54 @@ class CollectionInputFilter extends InputFilter
...
+     * @psalm-suppress NonInvariantDocblockPropertyType
+     */
+    protected $invalidInputs;
+
+    /**
+     * @var array<array-key, list<string>>|null
+     * @psalm-suppress NonInvariantDocblockPropertyType
+     */
+    protected $validationGroup;
+
```
gsteel: The collection is simply not compatible with a lot of the declared types in `InputFilterInterface` - I duplicated these inherited properties so that I could assign the more specific types. OFC I can revert and re-add the errors to the baseline

src/CollectionInputFilter.php
```diff
@@ -120,28 +141,36 @@ public function setCount($count)
...
+        /** @psalm-suppress DocblockTypeContradiction, RedundantConditionGivenDocblockType */
+        if (! is_array($data) && ! $data instanceof Traversable) {
             throw new Exception\InvalidArgumentException(sprintf(
                 '%s expects an array or Traversable collection; invalid collection of type %s provided',
                 __METHOD__,
                 is_object($data) ? get_class($data) : gettype($data)
             ));
         }
 
+        $data = ArrayUtils::iteratorToArray($data);
```
gsteel: Calling `iteratorToArray` here makes `$this->data` the expected type - at this point, in a collection, $data is a list of data arrays or possibly traversable - it shouldn't make any difference as `parent::setData()` also calls `iteratorToArray` - nonetheless, this might be risky

src/Factory.php
```diff
@@ -457,14 +489,18 @@ protected function injectDefaultFilterAndValidatorChains(InputInterface $input)
     protected function injectFilterAndValidatorChainsWithPluginManagers(InputInterface $input)
     {
         if ($this->defaultFilterChain) {
-            $input->getFilterChain()
-                ? $input->getFilterChain()->setPluginManager($this->defaultFilterChain->getPluginManager())
+            $filterChain = $input->getFilterChain();
+            /** @psalm-suppress RedundantConditionGivenDocblockType, DocblockTypeContradiction */
+            $filterChain instanceof FilterChain
+                ? $filterChain->setPluginManager($this->defaultFilterChain->getPluginManager())
```
gsteel, line 484: It's not actually feasible for `$filterChain` to be null here, but it breaks a couple of mocks in tests that _do_ return null for `getFilterChain` - Same thing for validators from Line 500

src/InputFilterAbstractServiceFactory.php
```diff
@@ -89,13 +92,14 @@ protected function getInputFilterFactory(ContainerInterface $container)
...
-            ->getDefaultFilterChain()
-            ->setPluginManager($this->getFilterPluginManager($container));
-        $this->factory
-            ->getDefaultValidatorChain()
-            ->setPluginManager($this->getValidatorPluginManager($container));
+        $this->factory  = new Factory();
+        $filterChain    = $this->factory->getDefaultFilterChain();
+        $validatorChain = $this->factory->getDefaultValidatorChain();
+        assert($filterChain instanceof FilterChain);
+        assert($validatorChain instanceof ValidatorChain);
```
gsteel, line 99: New instances will have non-null chains but you can't psalm-assert on a constructor :(

src/InputFilterInterface.php
```diff
@@ -1,18 +1,47 @@
...
 namespace Laminas\InputFilter;
 
 use Countable;
+use Laminas\Filter\FilterChain;
+use Laminas\Filter\FilterInterface;
+use Laminas\Validator\ValidatorChain;
+use Laminas\Validator\ValidatorInterface;
 use Traversable;
 
+/**
```
gsteel, line 12: Hopefully these definitions will provide some handy inference

src/InputFilterProviderInterface.php
```diff
@@ -6,7 +6,7 @@ interface InputFilterProviderInterface
 {
     /**
      * Should return an array specification compatible with
-     * {@link Laminas\InputFilter\Factory::createInputFilter()}.
+     * {@link Factory::createInputFilter()}.
      *
      * @return array
```
gsteel: This should return a new psalm-type for `InputFilterSpecification`

src/UnfilteredDataInterface.php
```diff
@@ -8,12 +8,12 @@
...
 {
     /**
-     * @return array|object
+     * @return array<array-key, mixed>
      */
     public function getUnfilteredData();
 
     /**
-     * @param array|object $data
+     * @param array<array-key, mixed> $data
```
gsteel, line 16: Should probably be changed to `iterable<array-key, mixed>`

gsteel (commented):
I should probably add a psalm-type for InputFilterSpecification that an InputFilterProvider can return, and the Factory can use

gsteel:
Current psalm failures will be solved by https://github.com/laminas/laminas-validator/pull/128

src/BaseInputFilter.php
```diff
@@ -104,6 +107,8 @@ public function add($input, $name = null)
             return $this;
         }
 
+        /** @psalm-var string $name */
```
Ocramius: This looks forced: I suggest leaving it be, and expanding the baseline

src/BaseInputFilter.php
```diff
@@ -318,22 +325,25 @@ public function setValidationGroup($name)
...
                     // Recursively populate validation groups for sub input filters
-                    $this->inputs[$key]->setValidationGroup($value);
+                    $input->setValidationGroup($value);
                 }
             }
         } else {
             $inputs = func_get_args();
         }
 
+        /** @psalm-var list<string> $inputs */
```
Ocramius: IMO we shouldn't force this here: larger baseline is better in this case

src/BaseInputFilter.php
```diff
@@ -408,11 +418,13 @@ public function getValues()
         foreach ($inputs as $name) {
             $input = $this->inputs[$name];
 
-            if ($input instanceof InputFilterInterface) {
-                $values[$name] = $input->getValues();
-                continue;
-            }
-            $values[$name] = $input->getValue();
+            /** @psalm-var mixed $value */
```
Ocramius: We can skip this variable assignment - we already suppress the `MixedAssignment` below

src/BaseInputFilter.php
```diff
@@ -530,13 +545,15 @@ protected function populate()
...
 
+            /** @psalm-var mixed $value */
             $value = $this->data[$name];
 
             if ($input instanceof InputFilterInterface) {
                 // Fixes #159
                 if (! is_array($value) && ! $value instanceof Traversable) {
                     $value = [];
                 }
+                /** @psalm-var iterable<array-key, mixed> $value */
```
Ocramius: This shouldn't be needed: something wrong with inference.

src/BaseInputFilter.php
```diff
@@ -530,13 +545,15 @@ protected function populate()
                 continue;
             }
 
+            /** @psalm-var mixed $value */
             $value = $this->data[$name];
 
             if ($input instanceof InputFilterInterface) {
                 // Fixes #159
                 if (! is_array($value) && ! $value instanceof Traversable) {
```
Ocramius: Perhaps applying De Morgan's here could improve inference?

src/CollectionInputFilter.php
```diff
@@ -20,34 +21,54 @@ class CollectionInputFilter extends InputFilter
...
 
-    /** @var NotEmpty */
+    /** @var NotEmpty|null */
     protected $notEmptyValidator;
 
+    /**
+     * @var array<array-key, array<string, InputInterface|InputFilterInterface>>|null
+     * @psalm-suppress NonInvariantDocblockPropertyType
+     */
+    protected $validInputs;
```
Ocramius: These added properties are suspicious, especially if `protected` :|

src/CollectionInputFilter.php
```diff
@@ -20,34 +21,54 @@ class CollectionInputFilter extends InputFilter
...
+     */
+    protected $validationGroup;
+
     /**
      * Set the input filter to use when looping the data
      *
      * @param BaseInputFilter|array|Traversable $inputFilter
      * @throws Exception\RuntimeException
      * @return CollectionInputFilter
+     * @psalm-assert BaseInputFilter $this->inputFilter
```
Ocramius: This looks like `@psalm-out` rather than `@psalm-assert` - we don't use this method like an assertion.

src/CollectionInputFilter.php
```diff
@@ -120,28 +141,36 @@ public function setCount($count)
...
         }
 
         return $this->count;
     }
 
     /**
-     * {@inheritdoc}
+     * @param mixed $data
+     * @psalm-suppress ImplementedParamTypeMismatch
+     * @psalm-assert array<array-key, mixed> $this->data
```
Ocramius: This would be picked up as redundant assertion, if used in a context where the given `$data` is already of the right shape 🤔

src/CollectionInputFilter.php
```diff
@@ -120,28 +141,36 @@ public function setCount($count)
...
                 __METHOD__,
                 is_object($data) ? get_class($data) : gettype($data)
             ));
         }
 
+        $data = ArrayUtils::iteratorToArray($data);
+
         $this->setUnfilteredData($data);
 
+        /** @psalm-var mixed $item */
```
Ocramius: Probably better to suppress the mixed assignment, than to add an `@var` block (which prevents inference)

src/Factory.php
```diff
@@ -43,7 +49,8 @@ public function __construct(?InputFilterPluginManager $inputFilterManager = null
     /**
      * Set default filter chain to use
      *
-     * @return Factory
+     * @psalm-assert FilterChain $this->defaultFilterChain
```
Ocramius: All over the patch, I'd say that  `@psalm-assert` is misused: let's not use it as a hack for internal state mutation.

src/Factory.php
```diff
@@ -238,6 +255,7 @@ public function createInput($inputSpecification)
                             is_object($value) ? get_class($value) : gettype($value)
                         ));
                     }
+                    /** @psalm-var iterable<array-key, FilterSpecification> $value */
```
Ocramius: Let's not force the value here - all over the patch, `@var` blocks need to be strongly reduced, preferring inference over declaration

Ocramius (changes requested):

src/BaseInputFilter.php
gsteel: Had to Google De Morgan - still none the wiser!

src/CollectionInputFilter.php
gsteel: I thought `@param-out` only worked for by ref arguments?

src/BaseInputFilter.php
gsteel: It's not wrong - just slighty 'off': `psalm: MixedArgumentTypeCoercion: Argument 1 of Laminas\InputFilter\InputFilterInterface::setData expects iterable<array-key, mixed>|null, parent type Traversable|array<array-key, mixed> provided`

I've baselined that

src/CollectionInputFilter.php
gsteel: Most of these asserts could be replaced with null checks where they're missing, and they're missing a fair bit - I favoured the annotation so I didn't have to change the code where it made sense. Would you like me to remove them and re-add to the baseline?

Ocramius: I would say that `@psalm-assert` is the wrong tool for the job here: adding `assert()` where necessary is a better solution.

Alternatively, keep them in baseline for now, but let's get rid of all `@psalm-assert` that are not strictly about asserting inputs

src/CollectionInputFilter.php
Ocramius: Same as all `@psalm-assert` introduced in this patch: let's remove them and replace them either with manual assertions, or with a larger baseline for now

src/CollectionInputFilter.php
Ocramius: Iterating twice on a potentially closed iterator is IMO worse - can we avoid adding this?

src/Input.php
```diff
@@ -227,6 +233,7 @@ public function breakOnFailure()
      */
     public function continueIfEmpty()
     {
+        /** @psalm-suppress DeprecatedProperty */
```
Ocramius: Let's keep these in baseline - they are supposed to be reduced/elided over time

src/Input.php
```diff
@@ -209,6 +214,7 @@ public function setFallbackValue($value)
      */
     public function allowEmpty()
     {
+        /** @psalm-suppress DeprecatedProperty */
```
Ocramius: Let's keep these in baseline - they are supposed to be reduced/elided over time

src/InputFilterAbstractServiceFactory.php
Ocramius: Even then, asserting shouldn't happen with `psalm-assert` as per discussions above :D

src/InputFilterInterface.php
```diff
@@ -1,18 +1,47 @@
-<?php
+<?php // phpcs:disable SlevomatCodingStandard.Namespaces.UnusedUses.UnusedUse
```
Ocramius: Can we perhaps have it on specific lines? :thinking:

src/InputFilterPluginManager.php
```diff
@@ -1,4 +1,4 @@
-<?php
+<?php //phpcs:disable WebimpressCodingStandard.PHP.CorrectClassNameCase.Invalid
```
Ocramius: Let's disable this at `phpcs.xml` level perhaps?

src/InputFilterPluginManagerFactory.php
```diff
@@ -1,4 +1,4 @@
-<?php
+<?php //phpcs:disable WebimpressCodingStandard.PHP.CorrectClassNameCase.Invalid
```
Ocramius: Let's exclude this at phpcs level, perhaps?

src/InputInterface.php
```diff
@@ -113,7 +113,7 @@ public function getValue();
     public function isValid();
 
     /**
-     * @return string[]
+     * @return array<array-key, string>
```
Ocramius, line 116: BTW, `array<string>` is the same as `array<array-key, string>`

src/OptionalInputFilter.php
```diff
@@ -5,22 +5,24 @@
...
      * Set data to use when validating and filtering
      *
-     * @param  iterable|mixed $data
-     *     must be a non-empty iterable in order trigger actual validation, else it is always valid
+     * @param iterable<array-key, mixed>|null $data must be a non-empty iterable in order trigger
+     *                                              actual validation, else it is always valid
+     * @return $this
      * @throws Exception\InvalidArgumentException
-     * @return InputFilterInterface
+     * @psalm-assert array<array-key, mixed> $this->data
```
Ocramius: To be dropped

src/OptionalInputFilter.php
```diff
@@ -5,22 +5,24 @@
...
- *
- * This is analog to {@see Laminas\InputFilter\Input} with the option ->setRequired(false)
+ * This is analog to {@see Input} with the option ->setRequired(false)
  */
 class OptionalInputFilter extends InputFilter
 {
     /**
      * Set data to use when validating and filtering
      *
-     * @param  iterable|mixed $data
```
Ocramius, line 16: Previously, this allowed `mixed` - now it only allows `iterable|null` :thinking:

test/FactoryTest.php
```diff
@@ -1115,22 +1106,19 @@ public function testFactoryCanCreateCollectionInputFilterWithRequiredMessage()
...
      * @param mixed $pluginValue
-     * @return MockObject|InputFilterPluginManager
+     * @return MockObject&InputFilterPluginManager
      */
     protected function createInputFilterPluginManagerMockForPlugin($pluginName, $pluginValue)
     {
-        /** @var InputFilterPluginManager|MockObject $pluginManager */
+        /** @var InputFilterPluginManager&MockObject $pluginManager */
         $pluginManager = $this->getMockBuilder(InputFilterPluginManager::class)
             ->disableOriginalConstructor()
```
Ocramius: This can be dropped, replaced by `$this->createMock(InputFilterPluginManager::class)`, without any `@var` declaration

Ocramius (changes requested):

src/InputFilterPluginManagerFactory.php
gsteel: I went for the inline suppression because the sniff is wrong - it's complaining that `Interop\Container\ContainerInterface` should actually be `interop\container\containerinterface` - this is probably something to do with Service Manager and that weird psr extension - I could replace with a doc block citing removal when Interop is finally dead? Otherwise, we would potentially be suppressing valid errors elsewhere no?

src/CollectionInputFilter.php
gsteel: parent::setData isn't called here therefore, if the collection is part of a containing input filter, the data is already an array, so iteratorToArray will be receiving an array anyhow - theoretically, this ensures that an iterable _is_ cast to an array if a collection is used as a top-level inputfilter or extended. I'll remove it anyway and baseline any additional errors

src/InputFilterInterface.php
gsteel: Yes - good idea, I'll `//phpcs: ignore` the offending lines

src/InputFilterPluginManagerFactory.php
Ocramius: Yeh, I'm aware of why this sniff was suppressed here. Thinking further about it, I think this is fine as-is, since the suppression only applies to factory and container definitions

src/OptionalInputFilter.php
gsteel: The annotation was just wrong imo - the change matches the interface and base abstract - these all stated out as `array|Traversable` and occasionally null too - I've changed this to `iterable<array-key, mixed>|null` - TBF, this could just be `iterable|null` couldn't it? The collection is still mixed, but even then `iterable|null` would be a more precise type - there's a suppression on collection for `ImplementedParamTypeMismatch` that should also be removed most likely.

Ocramius: > The annotation was just wrong imo

Works for me. 

 > TBF, this could just be `iterable|null` couldn't it?

yep

src/CollectionInputFilter.php
gsteel: @Ocramius - What shall I do about these added properties? Shall I bin them and re-add to the baseline?

Ocramius: Yes please: adding more `protected` properties is just more damage, IMO :|

gsteel:
I think that's all the points attended to @Ocramius - Thank you for the review and all your help :)

src/CollectionInputFilter.php
```diff
@@ -120,18 +124,20 @@ public function setCount($count)
     public function getCount()
     {
         if (null === $this->count) {
-            return count($this->data);
+            return $this->data !== null ? count($this->data) : 0;
```
Ocramius, line 127: Not important, but could've been `count($this->data ?? [])` :D

Ocramius (approved):
Awesome, thanks @gsteel!

PR #56, opened 2022-06-13
Improve type inference for plugin manager

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

If https://github.com/laminas/laminas-servicemanager/pull/137 gets merged, this is what I had in mind for `inputfilter`, `filter`, `validator` and `view` … and others??

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "53f38233dacf2b64b07daad58d574517",
+    "content-hash": "87202727d455197035daa0ce8b5a79e1",
     "packages": [
         {
             "name": "laminas/laminas-filter",
--- a/composer.lock
+++ b/composer.lock
@@ -87,16 +87,16 @@
         },
         {
             "name": "laminas/laminas-servicemanager",
-            "version": "3.11.2",
+            "version": "3.12.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/laminas/laminas-servicemanager.git",
-                "reference": "8a1f4d53ec93b2e18174f6f186922ef44d11a75a"
+                "reference": "05ac4b1fb1fe9333313eeafced9b6c7946589487"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/laminas/laminas-servicemanager/zipball/8a1f4d53ec93b2e18174f6f186922ef44d11a75a",
-                "reference": "8a1f4d53ec93b2e18174f6f186922ef44d11a75a",
+                "url": "https://api.github.com/repos/laminas/laminas-servicemanager/zipball/05ac4b1fb1fe9333313eeafced9b6c7946589487",
+                "reference": "05ac4b1fb1fe9333313eeafced9b6c7946589487",
                 "shasum": ""
             },
             "require": {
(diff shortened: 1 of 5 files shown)


Ocramius:
https://github.com/laminas/laminas-servicemanager/releases/tag/3.12.0

composer.lock
```diff
@@ -87,16 +87,16 @@
         },
         {
             "name": "laminas/laminas-servicemanager",
-            "version": "3.11.2",
+            "version": "3.12.0",
```
Ocramius, line 90: We should bump in `composer.json`

Ocramius (changes requested):

composer.lock
gsteel: Done… I figured the lock would cover it. Is it generally a good idea to bump minimums when adding psalm types not present in older versions?

Ocramius (approved):
Thanks @gsteel!

composer.lock
Ocramius: I think bumping is always a good idea, when it brings some goodies with it :D

FabianKoestring:
After this change we got a lot of PHPStan errors.

```php
Return type (array) of method Application\Form\Fieldset\File::getInputFilterSpecification() should be compatible with return type (*NEVER*) of method
         Laminas\InputFilter\InputFilterProviderInterface::getInputFilterSpecification()
```

`Application\Form\Fieldset\File::getInputFilterSpecification()` looks like this.
```php
public function getInputFilterSpecification(): array
{
    return [...];
}

**Comment by Ocramius** [comment]
@FabianKoestring can you please make a reproducer snippet on https://phpstan.org/try ?

**Comment by Ocramius** [comment]
Also, please do so in a new issue :)

**Comment by FabianKoestring** [comment]
> @FabianKoestring can you please make a reproducer snippet on https://phpstan.org/try ?

I would do that but I don't know how to set dependencies (to `laminas/laminas-inpufilter`) on https://phpstan.org/try. The problem occurs from version 2.16.0 on.

**Comment by Ocramius** [comment]
@FabianKoestring copy in all symbols that affect your issue, and reduce the example to the minimum size to show the issue.

**Comment by FabianKoestring** [comment]
> @FabianKoestring copy in all symbols that affect your issue, and reduce the example to the minimum size to show the issue.

👍 Look at error on line 62. https://phpstan.org/r/91e0e318-0ada-4777-9e50-983a0d7678a1

> Return type (CollectionSpecification|InputFilterSpecification) of method HelloWorld::getInputFilterSpecification() should be compatible with return type (*NEVER*) of method InputFilterProviderInterface::getInputFilterSpecification()

**Comment by Ocramius** [comment]
Simplifying your example further, it looks like phpstan doesn't understand this syntax:

```
array{
     type?: class-string<InputFilterInterface>|string,
}&array<array-key, InputSpecification>
```

See https://phpstan.org/r/6e758234-f4ef-4ff6-8519-919f3c9297e4

I would report an issue in upstream `phpstan/phpstan-src`

FabianKoestring:
@Ocramius https://github.com/phpstan/phpstan/issues/7475

froschdesign:
@gsteel 
Please check https://github.com/phpstan/phpstan/issues/7475#issuecomment-1155259861

gsteel:
The problem with the `InputFilterSpecification` type is that is seems valid to me. From what I can tell, the factory accepts `[?'type' => 'FQCN or alias']` and also the array will contain a list of either Inputs or InputFilters or specification arrays. The keys of the list can be strings or integers - string keys are used as the input name (This is the assumed behaviour) and integer keys are replaced with the 'name' from the input spec (value).

At the expense of allowing integer keys, the type could be `array{type?: whatever}&array<string, InputSpec|InstanceType>`

It currently seems to me that the intersection type is the right thing but I'd gladly learn of a better/more precise way to represent this…

PR #60, opened 2022-07-06
Regression for input filter collections

### LINKED ISSUE #58: [BUG] Newly introduced string check for validation groups in BaseInputFilter is causing problems for CollectionFieldset with integers as Id's
Since 2.16.0 there's an added string check. This check throws an exception when applying a validation group that contains integers instead of strings. The problem will occur when using a CollectionFieldset with entities that have integers as identifiers.

```
// BaseInputFilter.php
protected function validateValidationGroup(array $inputs)
  {
      foreach ($inputs as $name) {
          if (! is_string($name) || ! array_key_exists($name, $this->inputs)) {
              throw new Exception\InvalidArgumentException(sprintf(
                  'setValidationGroup() expects a list of valid input names; "%s" was not found',
                  (string) $name
              ));
          }
      }
  }
```

Solution could be to either remove the `!is_string()` check or to add an additional `is_int()` check as well `! (is_string($name) || is_int($name)) `

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes/no
| QA            | yes/no

### Description

Reference #58 

Validation groups appear to make sense for a regular "top-level" input filters and they also appear to work as advertised.

Applying validation groups to collection input filters doesn't appear to work unless I'm missing something… I've tried the tests here back to 2.15 with the same results.

Back in 2013, there was some `array_fill` shenanigans going on with the validation group, I assume to get around the problem where you have to repeatedly specify the grouped inputs per iteration of the collection, i.e. in a top-level input filter:

```php
$inputFilter->setValidationGroup([
    'collectionName' => [
        ['input1', 'input2'],
        ['input1', 'input2'],
        ['input1', 'input2'],
        // ... an so on, depending on the value of "count"
    ],
]);
```

FTR, I'm not expecting this to be merged, but it looks like there's a lot of reflection and mocking going on in the relevant tests here, hence the separate test cases - I'm not convinced the existing tests are covering the expected behaviour and when it comes to collections,I'm not even sure what the expected behaviour is - also, CI is broken in `laminas-form` for recent versions of input filter so there's definitely something broken…

### DIFF
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -497,7 +497,7 @@ class BaseInputFilter implements
     protected function validateValidationGroup(array $inputs)
     {
         foreach ($inputs as $name) {
-            if (! is_string($name) || ! array_key_exists($name, $this->inputs)) {
+            if (! array_key_exists($name, $this->inputs)) {
                 throw new Exception\InvalidArgumentException(sprintf(
                     'setValidationGroup() expects a list of valid input names; "%s" was not found',
                     (string) $name
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -21,7 +21,6 @@ use function gettype;
 use function is_array;
 use function is_int;
 use function is_object;
-use function is_string;
 use function sprintf;
 
 class BaseInputFilter implements
--- a/test/ValidationGroup/InputFilterCollectionsValidationGroupTest.php
+++ b/test/ValidationGroup/InputFilterCollectionsValidationGroupTest.php
@@ -9,6 +9,8 @@ use Laminas\InputFilter\Input;
 use Laminas\InputFilter\InputFilter;
 use PHPUnit\Framework\TestCase;
 
+use const PHP_VERSION_ID;
+
 final class InputFilterCollectionsValidationGroupTest extends TestCase
 {
     private InputFilter $inputFilter;
@@ -127,6 +129,7 @@ final class InputFilterCollectionsValidationGroupTest extends TestCase
         $this->setCollectionCount($count);
         $collection = $this->inputFilter->get('stuff');
         self::assertInstanceOf(CollectionInputFilter::class, $collection);
+        /** @psalm-suppress InvalidArgument */
         $collection->setValidationGroup([
             0 => 'first',
             1 => 'second',
@@ -147,7 +150,7 @@ final class InputFilterCollectionsValidationGroupTest extends TestCase
     }
 
     /**
-     * This test fails because of an undefined offset - the validation group must be set for elements 0 through 3
+     * This test documents existing behaviour - the validation group must be set for elements 0 through 3
      *
      * @dataProvider collectionCountProvider
      */
@@ -157,6 +160,7 @@ final class InputFilterCollectionsValidationGroupTest extends TestCase
         $collection = $this->inputFilter->get('stuff');
         self::assertInstanceOf(CollectionInputFilter::class, $collection);
 
+        /** @psalm-suppress InvalidArgument */
         $collection->setValidationGroup([
             0 => 'first',
         ]);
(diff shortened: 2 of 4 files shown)


test/ValidationGroup/InputFilterCollectionsValidationGroupTest.php
```diff
@@ -0,0 +1,200 @@
...
+        ]);
+
+        self::assertTrue($this->inputFilter->isValid());
+    }
+
+    /** @dataProvider collectionCountProvider */
+    public function testValidationGroupViaTopLevelInputFilter(?int $count): void
+    {
+        $this->setCollectionCount($count);
+        $this->inputFilter->setValidationGroup([
```
boesing, line 193: psalm is right here, this supposed to be:

```
[
   0 => ['first'],
   1 => ['second'],
   // etc.
]
```

Or am I missing something? Did this test pass in older releases? 🤔

boesing:
Should we retarget this against 2.16.x and then upmerge?

Ocramius:
@boesing this is only a test, not a bug fix though?

If it includes a fix, we should merge to current stable 👍

test/ValidationGroup/InputFilterCollectionsValidationGroupTest.php
gsteel: Psalm is currently wrong - I updated all the doc blocks based on the understanding that `validateValidationGroups()` only accepted `list<string>`, after all, calls to inputfilter::get() expect a string argument and everything is documented as having string keys…

At the moment, I can't figure out what the expected argument to `setValidationGroup` should be when you want to target a collection nested inside an input filter.

I think we need to update/add a test case that clearly shows the expected behaviour of `setValidationGroup` when targeting collections and then update the doc blocks to match and revert the `is_string` and possibly other `(string) $key` casts added in 2.16

FWIW, the tests added in this pull fail on 2.15 too so I can't currently verify that 2.16 broke anything - hence the need for additional test cases - perhaps @wilcol can chime in with some help on the expected behaviour.

boesing: > At the moment, I can't figure out what the expected argument to setValidationGroup should be when you want to target a collection nested inside an input filter.

It is always following keys.

```php
$inputFilter->setValidationGroup([
    'nameOfCollectionKey' => [
        /** This is nasty as when I had a look in the test cases from you with debugger enabled, a validation group for a collection expects every item of the collection has its own entry within this array, identified by the key */
]);
```

https://github.com/laminas/laminas-inputfilter/blob/8c663d35926f8276b4bf1a2c571310eb285f80cb/src/CollectionInputFilter.php#L234

gsteel: 👆 Yes… That's the conclusion I came to when reading the code - there's very little defence here - I'm getting missing offsets and strings being cast to arrays when you muck about with with the argument given to `setValidationGroup` - The tests I've submitted show off some of this, but it's all equally "broken" on 2.14 & 2.15 prior to the contested change in 2.16 - Documentation on collections and input filters also appears to be missing

src/BaseInputFilter.php
```diff
@@ -322,7 +322,7 @@ public function setValidationGroup($name)
         if (is_array($name)) {
             $inputs = [];
             foreach ($name as $key => $value) {
-                if (! $this->has((string) $key)) {
+                if (! $this->has($key)) {
```
Ocramius: Hmm, careful with this: integer keys are auto-cast: https://3v4l.org/FHqha

gsteel: Looking specifically at the causes for failures in `laminas-form`, I can confirm that either way, this has no effect. It's the _(now removed)_ call to `is_string` in `validateValidationGroup()` that's causing the test failures there. In this case, the cast is irrelevant because `has()` just simply calls `array_key_exists` which performs the same auto-cast. I'll revert so it's not confused with the actual problem 👍

Ocramius (approved):
LGTM, but probably need @wilcol's feedback

Ocramius:
:ship: meanwhile, thanks @gsteel!

PR #66, opened 2022-07-20
Remove dependency on prophecy

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

- Alters tests to use PHPUnit mocks enabling the removal of 2 dev dependencies
- Updates locked dependencies
- Bumps dev dependencies to most recent versions
- Expands psalm baseline with newly discovered issues, removing those solved by using phpunit mocks

There is 1 failing test case that should pass IMO - the previous prophecy constraint did not assert that `init()` is called only once for input filters. It is called twice because `init` is called from `validate` in input filter plugin manager, and `validate` is called both in `setService` and in `get` of `AbstractPluginManager`. I'm not convinced this is desirable behaviour, regardless, it can't be fixed here.

### DIFF
--- a/test/FactoryTest.php
+++ b/test/FactoryTest.php
@@ -22,20 +22,17 @@ use Laminas\Validator;
 use LaminasTest\InputFilter\TestAsset\CustomInput;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
-use Prophecy\Argument;
-use Prophecy\PhpUnit\ProphecyTrait;
-use ReflectionProperty;
 
 use function count;
 use function sprintf;
+
+// phpcs:ignore
 
 /**
  * @covers \Laminas\InputFilter\Factory
  */
 class FactoryTest extends TestCase
 {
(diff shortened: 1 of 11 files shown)


test/FactoryTest.php
```diff
@@ -22,20 +22,17 @@
...
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
-use Prophecy\Argument;
-use Prophecy\PhpUnit\ProphecyTrait;
-use ReflectionProperty;
 
 use function count;
 use function sprintf;
 
+// phpcs:ignore
```
Ocramius: Hmm? :thinking:

Ocramius (approved):

test/FactoryTest.php
gsteel: Missed that… Sorry. PHPStorm does that to Interop **every time** imports change 🙄

Ocramius:
Thanks @gsteel!

PR #67, opened 2022-07-20
Apply PHP 7.4 syntax and typed property

samsonasik:
Signed-off-by: Abdul Malik Ikhsan <samsonasik@gmail.com>

|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Since composer.json require php 7.4, php 7.4 syntax can be applied with typed properties. 

- For `final class`, typed properties for protected is allowed if no inherit of parent or used by its final class parent.
- Otherwise, update only private properties.

### DIFF
--- a/src/FileInput.php
+++ b/src/FileInput.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Laminas\InputFilter;
 
+use Laminas\InputFilter\FileInput\FileInputDecoratorInterface;
 use Psr\Http\Message\UploadedFileInterface;
 
 use function is_array;
@@ -31,7 +32,7 @@ class FileInput extends Input
     /** @var bool */
     protected $autoPrependUploadValidator = true;
 
-    private ?\Laminas\InputFilter\FileInput\FileInputDecoratorInterface $implementation = null;
+    private ?FileInputDecoratorInterface $implementation = null;
 
     /**
      * @inheritDoc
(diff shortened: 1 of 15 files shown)


samsonasik:
All green 🎉

src/FileInput.php
```diff
@@ -136,7 +136,9 @@ public function isValid($context = null)
             return true;
         }
 
-        return $this->implementation->isValid($context);
+        /** @var FileInputDecoratorInterface $implementation */
+        $implementation = $this->implementation;
+        return $implementation->isValid($context);
```
gsteel: Nitpick, but this could be `assert($this->implementation !== null);`

src/FileInput.php
```diff
@@ -136,7 +136,9 @@ public function isValid($context = null)
             return true;
         }
 
-        return $this->implementation->isValid($context);
+        /** @var FileInputDecoratorInterface $implementation */
```
Ocramius: This should be an assertion: `assert(... !== null)`

Ocramius (approved):

src/FileInput.php
samsonasik: updated https://github.com/laminas/laminas-inputfilter/pull/67/commits/987edc5398e16bda34ea0c5ea4b1f288a2f90e63

src/FileInput.php
samsonasik: updated https://github.com/laminas/laminas-inputfilter/pull/67/commits/987edc5398e16bda34ea0c5ea4b1f288a2f90e63

Ocramius:
Thanks @samsonasik!

Issue #34, opened 2021-06-17
Deprecated properties in docs


rarog:
### Bug Report

Documentation lists deprecated properties.

|    Q        |   A
|------------ | ------
| Version(s)  | 2.12.0

#### Summary

https://docs.laminas.dev/laminas-inputfilter/specs/ lists following

```php
                'allow_empty' => false,
                'continue_if_empty' => false,
```
Both properties are deprecated since 2.4.8

froschdesign:
And also here:

* https://docs.laminas.dev/laminas-inputfilter/application-integration/usage-in-a-laminas-mvc-application/
* https://docs.laminas.dev/laminas-inputfilter/application-integration/usage-in-a-mezzio-application/

Tasks:

* Update related pages with alternative usage
* Create a migration guide

froschdesign:
The related properties and methods are no longer deprecated:

* https://github.com/laminas/laminas-inputfilter/pull/77
* https://github.com/laminas/laminas-inputfilter/releases/tag/2.22.1

Issue #10, opened 2019-12-31
Feature/required validator


weierophinney:
These allows manage the required message like any other ValidatorInterface element.

Note: While you can add the validator manually the flag `setRequired` must be set too

``` php
  $input = new Input();
  $input->setRequired(true);empty value instead *not set*.
  $input->getValidatorChain()->attach(
      new Zend\InputFilter\Validator\Required(),
      true                             // break chain on failure

  );
```

``` php
  $inputSpecification = [
    'required'   => true,
    'validators' => [
      [
        'break_chain_on_failure' => true,
        'name'                   => 'Zend\\InputFilter\\Validator\\Required',
      ],
    ],
  ];
```

Note: `setRequired(false)` may not be enough and you will need remove the validator from the chain.

---

Originally posted by @Maks3w at https://github.com/zendframework/zend-inputfilter/pull/62

weierophinney:
Inject a validator in the chain make difficult reset the input state for to validate different values or transititions set <=> not set


---

Originally posted by @Maks3w at https://github.com/zendframework/zend-inputfilter/pull/62#issuecomment-137586919

weierophinney:
I haven't tried running this yet, but I like the idea.

Why must "the flag setRequired" be set when adding the validator manually? Wouldn't it be better to just deprecate the flag?


---

Originally posted by @larsnystrom at https://github.com/zendframework/zend-inputfilter/pull/62#issuecomment-147069548

weierophinney:
Because the validator does not validate the input value (https://github.com/zendframework/zend-inputfilter/pull/62/files#diff-5e41c64803af5c0ad8ad84c3e25c834cR412)

I didn't resolve how to make to replace the whole required funcionallity with the validator.

This is needed too https://github.com/zendframework/zend-inputfilter/blob/master/src/BaseInputFilter.php#L253


---

Originally posted by @Maks3w at https://github.com/zendframework/zend-inputfilter/pull/62#issuecomment-147070015

weierophinney:
Hmm, yeah that's true. Do you think it would be possible to use the $data array as $context instead of the raw values? That would enable the Required-validator to check whether a key exists without having a reference to the input object. 

You would have to inject the input's name in the Required-validator's constructor as well, so the validator knew which key to look for. 

What do you think?


---

Originally posted by @larsnystrom at https://github.com/zendframework/zend-inputfilter/pull/62#issuecomment-147071318

weierophinney:
[...]

PR #77, opened 2022-10-21
Removes deprecation messages which lead to irritation because there are no alternative paths for implementation

froschdesign:
The current deprecations of methods and properties cannot be resolved or bypassed by other options or workarounds. The `isValid` method of the `Input` class uses and needs these deprecated methods and properties:

https://github.com/laminas/laminas-inputfilter/blob/2e9a2f2b867eb423338a9948ae0027ec4e556378/src/Input.php#L370-L405

https://github.com/laminas/laminas-inputfilter/blob/2e9a2f2b867eb423338a9948ae0027ec4e556378/src/Input.php#L20-L32

https://github.com/laminas/laminas-inputfilter/blob/2e9a2f2b867eb423338a9948ae0027ec4e556378/src/Input.php#L46-L51

An alternative solution or way was never provided and in its current state, it would not work. Therefore the deprecation was too early and all discussions and developments in this direction have never been completed or are very old.

See:

- #10 
- #7
- #34 

As long as we do not have an alternative solution, the current deprecation messages are confusing.

### DIFF
--- a/src/EmptyContextInterface.php
+++ b/src/EmptyContextInterface.php
@@ -4,22 +4,15 @@ declare(strict_types=1);
 
 namespace Laminas\InputFilter;
 
-/**
- * @deprecated 2.4.8 Add Laminas\Validator\NotEmpty validator to the ValidatorChain.
- */
 interface EmptyContextInterface
 {
     /**
-     * @deprecated 2.4.8 Add Laminas\Validator\NotEmpty validator to the ValidatorChain and set this to `true`.
-     *
      * @param bool $continueIfEmpty
      * @return self
      */
     public function setContinueIfEmpty($continueIfEmpty);
 
     /**
-     * @deprecated 2.4.8 Add Laminas\Validator\NotEmpty validator to the ValidatorChain. Should always return `true`.
-     *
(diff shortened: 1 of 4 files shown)


Ocramius (approved):
Fine by me. Need another opinion from who introduced this deprecation

gsteel (approved):
I agree with the decision to un-deprecate this stuff. That said, I can see _why_ it was deprecated as the "is it empty/required" dance is convoluted to say the least and the magic of auto injecting the NotEmpty validator is questionable.

IMO it's worth raising an issue that it all needs sorting out for the next major.

Does removing the deprecations alter the baseline?

internalsystemerror (approved):
No objections here. When/if a solution is found, is there a chance these deprecations would be readded as they are?

Ocramius:
Ok, let's just re-generate the baseline, and then this can be merged.

Can't do myself: don't have docker on this PC today :D

Ocramius:
Merging as-is here: can't re-generate baseline ATM, but it's a minor issue.

Ocramius:
Thanks @froschdesign!

PR #78, opened 2022-11-02
Use typed properties for ValidatorChain & FilterChain in Input

pascalheidmann:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This PR will add property types for base `Input` class for `$validatorChain` and `$filterChain` properties. As those could be accessed in derived classes directly and not via their appropriate getters/setters this is breaking.
The purpose of this PR is to remove confusion for phpstan (not sure about psalm) as it will detect a `nullable` property as technically it is `null` as long it isn't set or initialized via `getValidatorChain()`/`getFilterChain()`.

### DIFF
--- a/src/Input.php
+++ b/src/Input.php
@@ -37,8 +37,8 @@ class Input implements
     /** @var string|null */
     protected $errorMessage;
 
-    /** @var null|FilterChain */
-    protected $filterChain;
+    /** @var FilterChain */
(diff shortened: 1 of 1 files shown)


Ocramius:
This can't really be done, as it is a BC break for child classes.

PR #84, opened 2023-02-01
Hotfix/return psalm types

gromson:
`@return` annotation specifying psalm types produces false errors for other static analysis tools, particularly for PHPStan. Therefore, returning psalm types specified with `@psalm-return` and native PHP types provided for `@return` annotation.

### DIFF
--- a/src/ConfigProvider.php
+++ b/src/ConfigProvider.php
@@ -17,7 +17,7 @@ class ConfigProvider
      *     input_filters: ServiceManagerConfigurationType,
      * }
      */
-    public function __invoke()
+    public function __invoke(): array
     {
         return [
             'dependencies'  => $this->getDependencyConfig(),
@@ -28,9 +28,10 @@ class ConfigProvider
     /**
      * Return dependency mappings for this component.
      *
-     * @return ServiceManagerConfigurationType
+     * @psalm-return ServiceManagerConfigurationType
+     * @return array
      */
-    public function getDependencyConfig()
+    public function getDependencyConfig(): array
     {
         return [
             'aliases'   => [
@@ -48,9 +49,10 @@ class ConfigProvider
     /**
      * Get input filter configuration
      *
-     * @return ServiceManagerConfigurationType
+     * @psalm-return ServiceManagerConfigurationType
+     * @return array
      */
-    public function getInputFilterConfig()
+    public function getInputFilterConfig(): array
     {
         return [
             'abstract_factories' => [
--- a/src/InputFilterProviderInterface.php
+++ b/src/InputFilterProviderInterface.php
@@ -14,7 +14,8 @@ interface InputFilterProviderInterface
      * Should return an array specification compatible with
      * {@link Factory::createInputFilter()}.
      *
-     * @return InputFilterSpecification|CollectionSpecification
+     * @psalm-return InputFilterSpecification|CollectionSpecification
+     * @return array
      */
     public function getInputFilterSpecification();
 }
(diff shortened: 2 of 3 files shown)


src/ConfigProvider.php
```diff
@@ -17,7 +17,7 @@ class ConfigProvider
      *     input_filters: ServiceManagerConfigurationType,
      * }
      */
-    public function __invoke()
+    public function __invoke(): array
```
Ocramius, line 20: This cannot be added as a native type, due to this class not being declared `final`

src/ConfigProvider.php
```diff
@@ -28,9 +28,10 @@ public function __invoke()
     /**
      * Return dependency mappings for this component.
      *
-     * @return ServiceManagerConfigurationType
+     * @psalm-return ServiceManagerConfigurationType
+     * @return array
```
Ocramius, line 32: This seems incorrect: returning `ServiceManagerConfigurationType` is indeed correct.

src/ConfigProvider.php
```diff
@@ -28,9 +28,10 @@ public function __invoke()
...
      */
-    public function getDependencyConfig()
+    public function getDependencyConfig(): array
```
Ocramius, line 34: This cannot be added as a native type, due to this class not being declared `final`

src/ConfigProvider.php
```diff
@@ -48,9 +49,10 @@ public function getDependencyConfig()
     /**
      * Get input filter configuration
      *
-     * @return ServiceManagerConfigurationType
+     * @psalm-return ServiceManagerConfigurationType
+     * @return array
      */
-    public function getInputFilterConfig()
+    public function getInputFilterConfig(): array
```
Ocramius, line 55: This cannot be added as a native type, due to this class not being declared `final`

Ocramius (changes requested):
This patch seems to be invalid.

Verified on the PHPStan playground to be working: https://phpstan.org/r/3d456956-4c62-4d8b-9167-f161281afc02

```php
<?php

/** @psalm-type ServiceManagerConfigurationType = array{foo: string}
interface ConfigInterface {}

/** @psalm-import-type ServiceManagerConfigurationType from ConfigInterface */
class ConfigProvider
{
    /** @return ServiceManagerConfigurationType */
    public function __invoke(): array
    {
        return [
            'foo' => 123, // should be a reported error
        ];
    }
}
```
```
Method ConfigProvider::__invoke() should return array{foo: string} but returns array{foo: 123}.
Offset 'foo' (string) does not accept type int.
```

gromson:
@Ocramius Please consider the following example (I'm only concerned about the client code issue on line 38)

https://phpstan.org/r/12d3d0f2-fe29-4ccb-bbc1-4cb4a6dfcc1f

This can be solved if  `InputFilterProviderInterface::getInputFilterSpecification()` will have `@return` annotation that is not specific to a psalm type

https://phpstan.org/r/4e715832-28f6-41b4-b18a-d20d2751f1e1

Would you consider the updated pull request?

Ocramius:
To me this sounds like `&array<array-key, InputSpecification>` is not supported in PHPStan, which is unrelated to having precise types declared here :thinking:

gromson:
This is indeed not supported in PHPStan but the point is that it's not valid to expect that, since `InputFilterSpecification` is not a real type/class/interface this is a Psalm-specific type alias and it's not intended to be fed to PHPStan. But because there are no fallbacks for analyzers other than Psalm the client code can not satisfy PHPStan because PHPStan can't figure out what a method from the 3rd party library returns. And this seems wrong to me.

All things considered, the suggestion of the substitution of `@return` annotations for psalm types to `@psalm-return` and providing `@return` annotations with native types seems reasonable to me.

Ocramius:
Ah yes, adding `@return` is acceptable 👍

Do send a new patch please

PR #85, opened 2023-02-01
Use `@psalm-return` annotations together with `@return` to improve compatibility with PHPStan static analysis

gromson:
The result of the discussion in https://github.com/laminas/laminas-inputfilter/pull/84


Ocramius (approved):
Thanks @gromson!

PR #90, opened 2023-05-24
Remove support for PHP 8.0, Upgrade PHPUnit to 10.1.x

gsteel:
Closes #86 

This is quite a big patch, but the tests and data providers are hellishly complex, so making them all static was a bit messy.

There's loads of inheritance going on in the tests which really needs to be refactored.


Ocramius (approved):
Thanks @gsteel!

PR #91, opened 2023-05-24
Improve consumer type inference with InputFilter templates

gsteel:
|    Q          |   A
|-------------- | ------
| RFC           | yes

### Description

Adds a template to InputFilterInterface to define the array shape of filtered values

The plan here is better inference when calling `InputFilterInterface::getValues()`. It might cause a lot of noise for users because psalm requires a template declaration for all inheritors, that said, it'd be useful to those who care.

Once merged, users will be able to define an input filter such as:

```php
/**
 * @psalm-type ValidPayload = array{
 *   inputA: non-empty-string,
 *   inputB: int,
 *   inputC: DateTimeImmutable,
 * @extends InputFilter<ValidPayload>
 */
final class MyInputFilter extends InputFilter
{
  public function init(): void
  {
    // Set up inputs according to desired constraints…
  }
}
```

… and psalm will infer the types for whatever comes out of `$inputFilter->getValues()` such as:

```php
$values = $inputFilter->getValues();

printf('The Date is %s', $values['inputC']->format('Y-m-d'));
```

As part of wrestling with Psalm on this patch, it's worth mentioning that all **input** names can now be `array-key` as opposed to `string`. This is not really related to the patch but something I had to address because `CollectionInputFilter` is effectively a nested list, it's actually a requirement to be able to fetch an input with something like `$inputFilter->get(1)`

This work has also uncovered at least one or two subtle bugs or ambiguous behaviour and fixes an issue where `InputFilter::add` will attempt to call `Input::merge` with an `InputFilter` as an argument

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -26,18 +26,12 @@
     </NonInvariantDocblockPropertyType>
   </file>
   <file src="src/BaseInputFilter.php">
-    <DocblockTypeContradiction>
-      <code><![CDATA[$input instanceof InputInterface && (empty($name) || is_int($name))]]></code>
-    </DocblockTypeContradiction>
     <InvalidReturnStatement>
-      <code>$messages</code>
+      <code>$values</code>
     </InvalidReturnStatement>
     <InvalidReturnType>
-      <code><![CDATA[array<string, array<array-key, string>>]]></code>
+      <code>TFilteredValues</code>
     </InvalidReturnType>
-    <MixedArgument>
-      <code>$input</code>
-    </MixedArgument>
     <MixedArgumentTypeCoercion>
       <code>$inputs</code>
     </MixedArgumentTypeCoercion>
@@ -52,9 +46,6 @@
     <PossiblyNullArrayOffset>
       <code><![CDATA[$this->inputs]]></code>
     </PossiblyNullArrayOffset>
-    <RedundantCastGivenDocblockType>
-      <code>(string) $name</code>
-    </RedundantCastGivenDocblockType>
     <TooManyArguments>
       <code>isValid</code>
       <code>isValid</code>
@@ -69,10 +60,6 @@
       <code>$name</code>
       <code>$name</code>
     </ImplementedParamTypeMismatch>
-    <ImplementedReturnTypeMismatch>
-      <code><![CDATA[array<array-key, array<string, array<array-key, string>>>]]></code>
-      <code><![CDATA[array<array-key, array<string, array<array-key, string>>>]]></code>
-    </ImplementedReturnTypeMismatch>
     <InvalidArgument>
       <code>$data</code>
     </InvalidArgument>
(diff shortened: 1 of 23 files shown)


src/BaseInputFilter.php
```diff
@@ -92,12 +96,15 @@ public function add($input, $name = null)
...
-        /** @psalm-suppress TypeDoesNotContainType */
         if ($input instanceof InputInterface && (empty($name) || is_int($name))) {
             $name = $input->getName();
         }
 
-        if (isset($this->inputs[$name]) && $this->inputs[$name] instanceof InputInterface) {
+        if (
+            isset($this->inputs[$name])
+            && $this->inputs[$name] instanceof InputInterface
+            && $input instanceof InputInterface
```
weierophinney, line 106: Why was this condition added?

weierophinney (changes requested):
Love the fact that you (a) added more tests to cover more scenarios, and (b) added the static analysis classes to validate the templates!

One big request, however: I think it would be fantastic to add some info to the docs about the Psalm shapes, and some examples of how they are used (these could be pulled from the static analysis cases you provide, or even just link to them). This will make it clear that we offer the feature; otherwise, it will only be discoverd by those diving into the source code.

psalm-baseline.xml
```diff
@@ -26,18 +26,12 @@
...
-    <DocblockTypeContradiction>
-      <code><![CDATA[$input instanceof InputInterface && (empty($name) || is_int($name))]]></code>
-    </DocblockTypeContradiction>
     <InvalidReturnStatement>
-      <code>$messages</code>
+      <code>$values</code>
     </InvalidReturnStatement>
     <InvalidReturnType>
-      <code><![CDATA[array<string, array<array-key, string>>]]></code>
+      <code>TFilteredValues</code>
```
Ocramius, line 33: Hmmmmm :thinking:

src/BaseInputFilter.php
gsteel: `InputFilter::add()` accepts `InputInterface|InputFilterInterface` - The condition was always there, asserting the left side was `InputInterface` but there was nothing to check the right side was the correct type - There's a test to cover it - effectively, we had the possibility of types smashing together with `InputInterface::merge(InputFilterInterface)` - I considered changing the original condition to just check the type equality but that would have broken BC - i.e. the existing behaviour would be a replace if the type was an `InputFilterInterface`

psalm-baseline.xml
gsteel: This is because inside `getValues()`, the return type is clearly `array<string, mixed>`, but because `TFilteredValues` is effectively `mixed` until a reasonable template is provided, psalm complains that `array<string, mixed>` is not `TFilteredValues` - it was choice between having the template unrestricted and everything working nicely for the user (and collections etc) vs a handful of `InvalidReturnType` issues as far as I could see. Any advice? I found that `TFilteredValues of array<array-key, mixed>` was more trouble than it was worth.

gsteel:
Added a first stab at documentation @weierophinney

weierophinney:
> Added a first stab at documentation

Perfect!

I'm approving now, but only because I don't fully understand Marco's concern (though your explanation makes sense to me).

weierophinney (approved):

psalm-baseline.xml
Ocramius: Works for me :)

Ocramius:
Thanks @gsteel!

PR #92, opened 2023-06-20
Using FQN instead of aliases when referencing the `NotEmpty` validator internally

vnccrl78:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### DIFF
--- a/src/Input.php
+++ b/src/Input.php
@@ -441,7 +441,7 @@ class Input implements
         $this->notEmptyValidator = true;
 
         if (class_exists(AbstractPluginManager::class)) {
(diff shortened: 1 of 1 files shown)


Slamdunk (approved):

Ocramius:
Thanks @vnccrl78!

boesing:
FYI: this could be a regression as upstream could've registered a factory for that service name.
Just a sidenote here, but messing with service names in minors is not a good idea. I cant even see a reason for this as of now, did you guys chat on slack and this was a result or why do we need this?

PR #97, opened 2024-01-09
Improve inference for `InputFilterPluginManager::get()`

gsteel:
Currently, `InputFilterPluginManager::get(FQCN::class)` yields the union of `InputFilterInterface|InputInterface`.

Ideally, this would be the given FQCN when it implements one of those interfaces.

Also, the `@method get` docblock is problematic as it trumps any user defined stubs and causes other problems that I can't remember!

Because the plugin manager here is guaranteed to return `InputFilterInterface|InputInterface` as opposed to `T|mixed`, this conflicts with `AbstractPluginManager::get()`.

Furthermore, I think psalm might have issues with `class-string<PsalmAliasOfUnion>`.

This patch provides correct inference for consumers when they provide a FQCN or string alias at the expense of 2 or 3 extra baselined issues that I believe are due to using a union type for `InstanceType` on the class level template.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -282,7 +282,7 @@
       <code>get</code>
     </MethodSignatureMustProvideReturnType>
     <MixedInferredReturnType>
-      <code><![CDATA[($name is class-string<T1> ? T1 : ($name is class-string<T2> ? T2 : T1|T2))]]></code>
+      <code><![CDATA[($name is class-string<T> ? T : InputInterface|InputFilterInterface)]]></code>
     </MixedInferredReturnType>
     <MixedReturnStatement>
       <code>parent::get($name, $options)</code>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -104,12 +104,6 @@
     </PossiblyUnusedReturnValue>
   </file>
   <file src="src/Factory.php">
-    <InvalidReturnStatement>
-      <code>$inputFilter</code>
-    </InvalidReturnStatement>
-    <InvalidReturnType>
-      <code>InputFilterInterface</code>
-    </InvalidReturnType>
     <MixedArgument>
       <code><![CDATA[$inputFilterSpecification['count']]]></code>
       <code><![CDATA[$inputFilterSpecification['input_filter']]]></code>
@@ -142,10 +136,6 @@
       <code>$value</code>
       <code>$value</code>
     </PossiblyInvalidArgument>
-    <PossiblyUndefinedMethod>
-      <code>add</code>
-      <code>add</code>
-    </PossiblyUndefinedMethod>
     <UnusedPsalmSuppress>
       <code>DeprecatedMethod</code>
       <code>DocblockTypeContradiction</code>
@@ -288,9 +278,6 @@
       <code>InputFilterPluginManager</code>
       <code>InputFilterPluginManager</code>
     </MethodSignatureMismatch>
-    <MethodSignatureMustProvideReturnType>
-      <code>Plu</code>
-    </MethodSignatureMustProvideReturnType>
     <NonInvariantDocblockPropertyType>
       <code>$factories</code>
     </NonInvariantDocblockPropertyType>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -278,6 +278,15 @@
       <code>InputFilterPluginManager</code>
       <code>InputFilterPluginManager</code>
     </MethodSignatureMismatch>
+    <MethodSignatureMustProvideReturnType>
+      <code>get</code>
+    </MethodSignatureMustProvideReturnType>
+    <MixedInferredReturnType>
+      <code><![CDATA[($name is class-string<T1> ? T1 : ($name is class-string<T2> ? T2 : T1|T2))]]></code>
+    </MixedInferredReturnType>
+    <MixedReturnStatement>
+      <code>parent::get($name, $options)</code>
+    </MixedReturnStatement>
     <NonInvariantDocblockPropertyType>
       <code>$factories</code>
     </NonInvariantDocblockPropertyType>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -282,7 +282,7 @@
       <code>get</code>
     </MethodSignatureMustProvideReturnType>
     <MixedInferredReturnType>
-      <code><![CDATA[($name is class-string<T> ? T : InputInterface|InputFilterInterface)]]></code>
+      <code><![CDATA[($name is class-string<InputInterface> ? T1 : ($name is class-string<InputFilterInterface> ? T2 : InputInterface|InputFilterInterface))]]></code>
     </MixedInferredReturnType>
     <MixedReturnStatement>
       <code>parent::get($name, $options)</code>
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -12,6 +12,7 @@ use Laminas\Validator\ValidatorChain;
 use Laminas\Validator\ValidatorInterface;
 use Traversable;
 
+use function assert;
 use function class_exists;
 use function get_debug_type;
 use function gettype;
@@ -317,6 +318,7 @@ class Factory
         }
 
         $inputFilter = $this->getInputFilterManager()->get($type);
+        assert($inputFilter instanceof InputFilterInterface); // As opposed to InputInterface
 
         if ($inputFilter instanceof CollectionInputFilter) {
             $inputFilter->setFactory($this);
--- a/src/InputFilterPluginManager.php
+++ b/src/InputFilterPluginManager.php
@@ -190,9 +190,12 @@ class InputFilterPluginManager extends AbstractPluginManager
 
     /**
      * @inheritDoc
-     * @template T of InputInterface|InputFilterInterface
-     * @param class-string<T>|string $name
-     * @return ($name is class-string<T> ? T : InputInterface|InputFilterInterface)
+     * phpcs:disable Generic.Files.LineLength.TooLong
+     * // Template constraint required or we get mixed added to output. Two templates because union does not work
+     * @template T1 of InputInterface
+     * @template T2 of InputFilterInterface
+     * @param class-string<T1>|class-string<T2>|string $name
+     * @return ($name is class-string<InputInterface> ? T1 : ($name is class-string<InputFilterInterface> ? T2 : InputInterface|InputFilterInterface))
      */
     public function get($name, ?array $options = null)
     {
--- a/src/InputFilterPluginManager.php
+++ b/src/InputFilterPluginManager.php
@@ -26,7 +26,6 @@ use function sprintf;
  * @psalm-import-type ServiceManagerConfiguration from ServiceManager
  * @template InstanceType of InputFilterInterface|InputInterface
  * @extends AbstractPluginManager<InstanceType>
- * @method InputFilterInterface|InputInterface get(string $name, ?array $options = null)
  */
 class InputFilterPluginManager extends AbstractPluginManager
 {
@@ -187,5 +186,18 @@ class InputFilterPluginManager extends AbstractPluginManager
         } catch (InvalidServiceException $e) {
             throw new Exception\RuntimeException($e->getMessage(), $e->getCode(), $e);
         }
+    }
+
+    /**
+     * @inheritDoc
+     * @template T1 of InputInterface
+     * @template T2 of InputFilterInterface
+     * @param class-string<T1>|class-string<T2>|string $name
+     * @return T1|T2
+     * @psalm-return ($name is class-string<T1> ? T1 : ($name is class-string<T2> ? T2 : T1|T2))
+     */
+    public function get($name, ?array $options = null)
+    {
+        return parent::get($name, $options);
     }
 }
--- a/src/InputFilterPluginManager.php
+++ b/src/InputFilterPluginManager.php
@@ -190,11 +190,9 @@ class InputFilterPluginManager extends AbstractPluginManager
 
     /**
      * @inheritDoc
-     * @template T1 of InputInterface
-     * @template T2 of InputFilterInterface
-     * @param class-string<T1>|class-string<T2>|string $name
-     * @return T1|T2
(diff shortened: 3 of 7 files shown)


src/Factory.php
```diff
@@ -317,6 +318,7 @@ public function createInputFilter($inputFilterSpecification)
         }
 
         $inputFilter = $this->getInputFilterManager()->get($type);
+        assert($inputFilter instanceof InputFilterInterface); // As opposed to InputInterface
```
Xerkus, line 321: How do we know it is not an InputInterface here? InputFilterManager allows both

gsteel: Because the `createInputFilter` method expects a spec (Array of type `InputFilterSpecification`). Here `$type` defaults to `InputFilter::class` and `InputFilterSpecification['type']` is `class-string<InputFilterInterface>`

Xerkus: I think this ought to be full check for input filter with meaningful exception, not just an assert since the method is user facing and there is no soft guard somewhere else.

src/InputFilterPluginManager.php
```diff
@@ -188,4 +187,17 @@ public function validatePlugin($plugin)
...
+     * @template T1 of InputInterface
+     * @template T2 of InputFilterInterface
+     * @param class-string<T1>|class-string<T2>|string $name
+     * @return T1|T2
+     * @psalm-return ($name is class-string<T1> ? T1 : ($name is class-string<T2> ? T2 : T1|T2))
+     */
+    public function get($name, ?array $options = null)
+    {
+        return parent::get($name, $options);
+    }
```
Ocramius: Isn't the entire signature the same as `@param class-string<T> @return T`? 

Overall OK with this, just seems redundant with the parent DIC declarations

Ocramius (approved):

src/InputFilterPluginManager.php
Xerkus: I am testing this locally. It seems it could be simplified.

Xerkus: I tried following:

```
    /**
     * @inheritDoc
     * @template T of InputInterface|InputFilterInterface
     * @param class-string<T>|string $name
     * @psalm-return ($name is class-string<T> ? T : T)
     */
    public function get($name, ?array $options = null)
```

Result is close:
```
ERROR: Trace - test/StaticAnalysis/InputFilterPluginManagerType.php:27:9 - $input: Laminas\InputFilter\Input (see https://psalm.dev/224)
        /** @psalm-trace $input */
        $input = $this->manager->get(Input::class);

ERROR: Trace - test/StaticAnalysis/InputFilterPluginManagerType.php:29:9 - $input: Laminas\InputFilter\InputInterface (see https://psalm.dev/224)
        /** @psalm-trace $input */
        $input = $this->manager->get('string');

ERROR: Trace - test/StaticAnalysis/InputFilterPluginManagerType.php:31:9 - $input: Laminas\InputFilter\InputInterface (see https://psalm.dev/224)
        /** @psalm-trace $input */
        $input = $this->manager->get('stdClass');

ERROR: Trace - test/StaticAnalysis/InputFilterPluginManagerType.php:33:9 - $input: LaminasTest\InputFilter\StaticAnalysis\InputFilterWithTemplatedValues (see https://psalm.dev/224)
        /** @psalm-trace $input */
        $input = $this->manager->get(InputFilterWithTemplatedValues::class);
```

For string parameters values are wrong.

However if I replace conditional fallback with explicit types it works out:

```
    /**
     * @inheritDoc
     * @template T of InputInterface|InputFilterInterface
     * @param class-string<T>|string $name
     * @psalm-return ($name is class-string<T> ? T : InputInterface|InputFilterInterface)
     */
    public function get($name, ?array $options = null)
```
```
ERROR: Trace - test/StaticAnalysis/InputFilterPluginManagerType.php:27:9 - $input: Laminas\InputFilter\Input (see https://psalm.dev/224)
        /** @psalm-trace $input */
        $input = $this->manager->get(Input::class);

ERROR: Trace - test/StaticAnalysis/InputFilterPluginManagerType.php:29:9 - $input: Laminas\InputFilter\InputFilterInterface|Laminas\InputFilter\InputInterface (see https://psalm.dev/224)
        /** @psalm-trace $input */
        $input = $this->manager->get('string');

ERROR: Trace - test/StaticAnalysis/InputFilterPluginManagerType.php:31:9 - $input: Laminas\InputFilter\InputFilterInterface|Laminas\InputFilter\InputInterface (see https://psalm.dev/224)
        /** @psalm-trace $input */
        $input = $this->manager->get('stdClass');

ERROR: Trace - test/StaticAnalysis/InputFilterPluginManagerType.php:33:9 - $input: LaminasTest\InputFilter\StaticAnalysis\InputFilterWithTemplatedValues (see https://psalm.dev/224)
        /** @psalm-trace $input */
        $input = $this->manager->get(InputFilterWithTemplatedValues::class);
```

Xerkus: Actually nevermind. This is wrong with both your version and my smiplification

```
ERROR: Trace - test/StaticAnalysis/InputFilterPluginManagerType.php:32:9 - $input: PHPUnit\Framework\TestCase (see https://psalm.dev/224)
        /** @psalm-trace $input */
        $input = $this->manager->get(TestCase::class);
```
I don't think conditional actually uses template there. It creates transient template instead used only for conditional.

gsteel: I was having problems with `class-string<InstanceType>` where `InstanceType` is the union, and similarly with `class-string<InputInterface|InputFilterInterface>` so that's why I went for `class-string<InputInterface>|class-string<InputFilterInterface>` - the simplified version is much better. I'll update the patch. Perhaps the cache was messing it up for me…

Xerkus: The problem is psalm both accepts wrong input and allows wrong input out. look at my last comment where I used TestCase as input

src/Factory.php
gsteel: An exception here would be a BC break though wouldn't it? The assertion will blow up in tests, it verifies the intended type for static analysis, and otherwise is a no-op for users

Xerkus: It wouldn't be. It would just be a fatal error if method not implemented or kinda work with anything as long as object has `add()` or `__call()`

Xerkus: This is more of a usability concern really. Exception would allow to say it needs to be input filter and get_debug_type was returned for the requested type `$type`.

src/InputFilterPluginManager.php
gsteel: Psalm is not wrong there because `'stdClass'` is a string, not a class-string, so it correctly yields the fallback and `InputFilterWithTemplatedValues` implements `InputFilterInterface` - it's not a TestCase, so it yields the FQCN which is exactly what we're after here. Or am I misunderstanding something?

Xerkus: TestCase is NOT input filter and it is NOT input either. Psalm is happy about telling returned type is TestCase

gsteel: Yeah, I just tested this. What you're saying is that Psalm will happily return `Anything` with `$manager->get(Anything::class)` - TBH, I thought that the restricted template `@template T of SpecificType` would prevent that, but it doesn't appear to.

Xerkus: I am trying alternatives. standby

Xerkus: Ok, so here is the results:
This is not a valid expression  `@template T of InputInterface|InputFilterInterface`. If I drop string from parameter I get:
```
ERROR: InvalidArgument - test/StaticAnalysis/InputFilterPluginManagerType.php:34:38 - Argument 1 of Laminas\InputFilter\InputFilterPluginManager::get expects class-string<Laminas\InputFilter\InputInterface>, but LaminasTest\InputFilter\StaticAnalysis\InputFilterWithTemplatedValues::class provided (see https://psalm.dev/004)
        $input = $this->manager->get(InputFilterWithTemplatedValues::class);
```
This indicates that template constraint is `InputInterface` and union part was ignored.

This is closer but has other issues:
```
    /**
     * @inheritDoc
     * @template T
     * @param class-string<T>|string $name
     * @return ($name is class-string<InputInterface>|class-string<InputFilterInterface> ? T : InputInterface|InputFilterInterface)
     */
```

Psalm then insists on throwing in mixed here
```
ERROR: MixedAssignment - src/Factory.php:174:13 - Unable to determine the type that $managerInstance is being assigned to (see https://psalm.dev/032)
            /** @psalm-trace $managerInstance */
            $managerInstance = $this->getInputFilterManager()->get($class);

  The type of $managerInstance is sourced from here - src/InputFilterPluginManager.php:194:16
     * @return ($name is class-string<InputInterface>|class-string<InputFilterInterface> ? T : InputInterface|InputFilterInterface)

ERROR: Trace - src/Factory.php:174:13 - $managerInstance: Laminas\InputFilter\InputFilterInterface|Laminas\InputFilter\InputInterface|mixed (see https://psalm.dev/224)
            /** @psalm-trace $managerInstance */
            $managerInstance = $this->getInputFilterManager()->get($class);


ERROR: MixedInferredReturnType - test/StaticAnalysis/InputFilterPluginManagerType.php:21:8 - Could not verify return type 'Laminas\InputFilter\InputFilterInterface|Laminas\InputFilter\InputInterface' for LaminasTest\InputFilter\StaticAnalysis\InputFilterPluginManagerType::getWillReturnAnInputOrInputFilterGivenAString (see https://psalm.dev/047)
    ): InputInterface|InputFilterInterface {


ERROR: MixedReturnStatement - test/StaticAnalysis/InputFilterPluginManagerType.php:22:16 - Possibly-mixed return value (see https://psalm.dev/138)
        return $this->manager->get($anyString);


ERROR: MixedReturnStatement - test/StaticAnalysis/InputFilterPluginManagerType.php:22:16 - Could not infer a return type (see https://psalm.dev/138)
        return $this->manager->get($anyString);


ERROR: Trace - test/StaticAnalysis/InputFilterPluginManagerType.php:28:9 - $input: Laminas\InputFilter\Input (see https://psalm.dev/224)
        /** @psalm-trace $input */
        $input = $this->manager->get(Input::class);


ERROR: Trace - test/StaticAnalysis/InputFilterPluginManagerType.php:30:9 - $input: Laminas\InputFilter\InputFilterInterface|Laminas\InputFilter\InputInterface (see https://psalm.dev/224)
        /** @psalm-trace $input */
        $input = $this->manager->get('string');


ERROR: Trace - test/StaticAnalysis/InputFilterPluginManagerType.php:32:9 - $input: Laminas\InputFilter\InputFilterInterface|Laminas\InputFilter\InputInterface (see https://psalm.dev/224)
        /** @psalm-trace $input */
        $input = $this->manager->get(TestCase::class);


ERROR: Trace - test/StaticAnalysis/InputFilterPluginManagerType.php:34:9 - $input: LaminasTest\InputFilter\StaticAnalysis\InputFilterWithTemplatedValues (see https://psalm.dev/224)
        /** @psalm-trace $input */
        $input = $this->manager->get(InputFilterWithTemplatedValues::class);

```

Xerkus: This is uncharted waters and I feel we are hitting at least couple edge case bugs here.

So, for current psalm version following works:
```
    /**
     * @inheritDoc
     * // Template constraint required or we get mixed added to output. Two templates because union does not work
     * @template T1 of InputInterface
     * @template T2 of InputFilterInterface
     * // string would allow anything and both templates currently receive the passed class ignoring constraint
     * // use both templates because of the first point.
     * @param class-string<T1>|class-string<T2>|string $name
     * // We can combine conditionals and use one of templates or swap templates and it will still work. Keep both as it will likely remain working in the future
     * @return ($name is class-string<InputInterface> ? T1 : ($name is class-string<InputFilterInterface> ? T2 : InputInterface|InputFilterInterface))
     */
```

Xerkus: @gsteel Psalm tests could be improved with https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-check-type to ensure we get expected types

I am sure it will not cause us any more grief on another lockfile maintenance PR :laughing:

Xerkus (changes requested):
psalm conditional does not work as expected. Working variant was found. It needs to be applied and checked.

src/InputFilterPluginManager.php
gsteel: @xerkus - just looking at this now and unfortunately, the string case union return type is messed up:

https://psalm.dev/r/84bf036569

Xerkus: https://psalm.dev/r/6e213fb5a6 Looks like a bug with `@psalm-check-type` not resolving FQCN against use statements and resolving against namespace for right side of union.

Xerkus (approved):

psalm-baseline.xml
```diff
@@ -289,8 +279,14 @@
       <code>InputFilterPluginManager</code>
     </MethodSignatureMismatch>
     <MethodSignatureMustProvideReturnType>
-      <code>Plu</code>
+      <code>get</code>
     </MethodSignatureMustProvideReturnType>
```
Xerkus, line 283: Why psalm complains about plugin manager get? Does it want simple typehinted return and conditional be in psalm-return?

Also what is the MethodSignatureMismatch just above that was previously baselined? Is it due to psr container aliasing shenanigans?

Xerkus: To answer it: some inherited weirdness going on.
```
ERROR: MethodSignatureMismatch - src/InputFilterPluginManager.php:30:7 - Method Laminas\ServiceManager\ServiceManager::has with return type '' is different to return type 'bool' of inherited method Psr\Container\ContainerInterface::has (see https://psalm.dev/042)
ERROR: MethodSignatureMismatch - src/InputFilterPluginManager.php:30:7 - Method Laminas\InputFilter\InputFilterPluginManager::get with return type '' is different to return type 'object' of inherited method Psr\Container\ContainerInterface::get (see https://psalm.dev/042)
```

gsteel: I think a lot of this will be solved with SMv4 when we can drop PSR container v1 and tighten up the templates and types in the plugin manager.

PR #99, opened 2024-02-01
Upgrade psalm to 5.21.x

gsteel:
Fixes various `RiskyTruthyFalsyComparison` issues.

Closes #98

### DIFF
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -96,7 +96,7 @@ class BaseInputFilter implements
             ));
         }
 
-        if ($input instanceof InputInterface && (empty($name) || is_int($name))) {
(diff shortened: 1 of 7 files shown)


src/BaseInputFilter.php
```diff
@@ -96,7 +96,7 @@ public function add($input, $name = null)
             ));
         }
 
-        if ($input instanceof InputInterface && (empty($name) || is_int($name))) {
+        if ($input instanceof InputInterface && ($name === null || $name === '' || is_int($name))) {
```
Xerkus, line 99: psalm essentially blacklists `empty()` it seems.

gsteel: Which is not such a bad thing 😁

Xerkus (approved):

PR #105, opened 2024-03-26
`ArrayInput`: Accept non-array input without throwing an exception

### LINKED ISSUE #104: Array Input has strange behaviour
I would expect that using an `ArrayInput` with `InputFilter` to fail validation when given non-array input, but instead an exception is thrown.

Given that the whole idea is filtering and validation of untrusted input, causing a crash in this situation seems odd to me.

```php
$inputFilter = new InputFilter();
$inputFilter->add([
    'type'       => ArrayInput::class,
    'validators' => [
        ['name' => NotEmptyValidator::class],
    ],
], 'myInput');

$inputFilter->setData(['myInput' => 'Whatever']); // Exception thrown here.
assert($inputFilter->isValid() === false);
```

gsteel:
### Description

Alters the behaviour of `ArrayInput` so that non-array input is deemed a validation failure instead of an exceptional condition.

Fixes #104

This could be considered a BC break, but it would _never_ have previously been possible for a non-array value to be validated/filtered here, so it's more likely that no-one ever used this input type without wrapping its execution in a try/catch block.

Assuming users are using this input type, and they _are_ wrapping with try, then the introduction of the IsArray validator will still provide graceful validation failure.

### DIFF
--- a/src/ArrayInput.php
+++ b/src/ArrayInput.php
@@ -4,60 +4,30 @@ declare(strict_types=1);
 
 namespace Laminas\InputFilter;
 
-use function gettype;
+use Laminas\Validator\IsArray;
+
+use function array_map;
+use function assert;
 use function is_array;
-use function sprintf;
 
 class ArrayInput extends Input
 {
-    /** @var array<array-key, mixed> */
-    protected $value = [];
-
-    /**
-     * @inheritDoc
-     * @param  array<array-key, mixed> $value
-     * @throws Exception\InvalidArgumentException
-     */
-    public function setValue($value)
-    {
-        if (! is_array($value)) {
-            throw new Exception\InvalidArgumentException(sprintf(
-                'Value must be an array, %s given.',
-                gettype($value)
-            ));
-        }
-        parent::setValue($value);
-
-        return $this;
-    }
-
-    /**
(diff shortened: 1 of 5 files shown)


src/ArrayInput.php
```diff
@@ -4,60 +4,30 @@
 
 namespace Laminas\InputFilter;
 
-use function gettype;
+use Laminas\Validator\IsArray;
```
Ocramius, line 7: Fancy new tech :P

Ocramius (approved):
LGTM :+

Ocramius:
Marked as bug + improvement on purpose, BTW: it changes dependencies, so it can't go to a patch version

gsteel:
Thankyou @Ocramius 🙏

src/ArrayInput.php
```diff
@@ -4,60 +4,30 @@
...
         }
-        parent::setValue($value);
-
-        return $this;
-    }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function resetValue()
```
kersysgediminas, line 37: @gsteel After removing this, default value is null now instead of empty array. It looks like a breaking change.

PR #113, opened 2024-07-12
Update `laminas-validator` to `2.61.0`

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

[2.61.0](https://github.com/laminas/laminas-validator/releases/tag/2.61.0) has a number of deprecations.

Tests adjusted to account for deprecations and those that cannot be dealt with without a BC break are base-lined

Minimum validator version requirement is bumped to 2.60 so that we can replace usage of the deprecated `Between` validator with the `NumberComparison` validator

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
         "laminas/laminas-filter": "^2.19",
         "laminas/laminas-servicemanager": "^3.21.0",
         "laminas/laminas-stdlib": "^3.0",
-        "laminas/laminas-validator": "^2.52"
+        "laminas/laminas-validator": "^2.60.0"
     },
     "require-dev": {
         "ext-json": "*",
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -35,7 +35,7 @@
         "laminas/laminas-filter": "^2.19",
         "laminas/laminas-servicemanager": "^3.21.0",
         "laminas/laminas-stdlib": "^3.0",
-        "laminas/laminas-validator": "^2.52"
+        "laminas/laminas-validator": "^2.60.0"
```
froschdesign, line 38: 2.60.0?

gsteel: It's the minimum required 🤷‍♂️

froschdesign (approved):
LGTM 👍🏻

PR #115, opened 2024-11-15
Add support for PHP 8.4

acelaya:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

This PR adds support for PHP 8.4 to this library.

Everything worked by simply adding `~8.4.0` to the supported PHP versions, but I had to bump the minimum required version of `laminas/laminas-stdlib` from `^3.0` to `^3.19`, otherwise tests fail with lowest dependencies in PHP 8.4

The change from `laminas/laminas-stdlib` required to avoid the deprecation error is this https://github.com/laminas/laminas-stdlib/pull/113/files#diff-1e1ee4f9152891e7f268ce1e744edb3426cb11cc7da5dc2033157cf7fef6b50aR45

### DIFF
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -1,6 +1,7 @@
 {
     "backwardCompatibilityCheck": true,
     "ignore_php_platform_requirements": {
-        "8.3": false
+        "8.3": false,
+        "8.4": true
     }
 }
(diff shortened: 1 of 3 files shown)


gsteel (approved):
LGTM! Thank you @acelaya

PR #124, opened 2025-05-11
Stop mocking Laminas\Validator\NotEmpty

ramchale:
Rasing draft PR to discuss how to tackle the tests that fail using a real NotEmpty instance.

Currently ArrayInputTest, HttpServerFileInputDecoratorTest, and PsrFileInputDecoratorTest all inherit from InputTest.

HttpServerFileInputDecoratorTest and PsrFileInputDecoratorTest both override`emptyValueProvider` with methods that provide values that do not fail validation as "NotEmpty". This causes `testDoNotInjectNotEmptyValidatorIfAnywhereInChain` to fail.

1st question is, should we remove the test inheriatance? Given there are several tests overridden so they can be skipped with `self::markTestSkipped('Test are not enabled in FileInputTest')`, this might be making things more difficult to maintain than have some duplicated logic. (We could also potentially use a trait to just depuplicate the ones that are actully common to all)

2nd question. Is there any intent to remove some or all of the chain of inheritence in the future that would negate the need for this test? The logic to automatically add a NotEmpty validator in Input, is overriden in FileInput to not run (which HttpServerFileInputDecorator and PsrFileInputDecorator inherit from).

Any other comments or suggestions on an alternative appraoch would also be appreciated

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

As of laminas-validator v3 NotEmpty is final and cannot be mocked by PHPUnit

### DIFF
--- a/test/ArrayInputTest.php
+++ b/test/ArrayInputTest.php
@@ -165,19 +165,6 @@ final class ArrayInputTest extends InputTest
         return parent::createValidatorChainMock($valueMap, $messages);
     }
 
-    protected function createNonEmptyValidatorMock(
-        bool $isValid,
-        mixed $value,
-        mixed $context = null,
-    ): NotEmptyValidator&MockObject {
-        // ArrayInput validates per each array value
-        if (is_array($value)) {
-            $value = current($value);
-        }
-
-        return parent::createNonEmptyValidatorMock($isValid, $value, $context);
-    }
-
     /** @return string[] */
     protected function getDummyValue(bool $raw = true)
     {
--- a/test/InputTest.php
+++ b/test/InputTest.php
@@ -267,14 +267,8 @@ class InputTest extends TestCase
             NotEmptyValidator::IS_EMPTY => "Custom message",
         ];
 
-        $notEmpty = $this->createMock(NotEmptyValidator::class);
-        $notEmpty->expects(self::once())
-            ->method('getOption')
-            ->with('messageTemplates')
-            ->willReturn($customMessage);
-
         $input->getValidatorChain()
-            ->attach($notEmpty);
+            ->attach(new NotEmptyValidator(['messages' => $customMessage]));
 
         self::assertFalse(
             $input->isValid(),
@@ -421,15 +415,15 @@ class InputTest extends TestCase
         $this->input->setRequired(true);
         $this->input->setValue($raw);
 
-        $notEmptyMock = $this->createNonEmptyValidatorMock(false, $raw);
+        $notEmptyValidator = new NotEmptyValidator();
 
         $validatorChain = $this->input->getValidatorChain();
-        $validatorChain->prependValidator($notEmptyMock);
+        $validatorChain->prependValidator($notEmptyValidator);
         self::assertFalse($this->input->isValid());
 
         $validators = $validatorChain->getValidators();
         self::assertEquals(1, count($validators));
-        self::assertEquals($notEmptyMock, $validators[0]['instance']);
+        self::assertEquals($notEmptyValidator, $validators[0]['instance']);
     }
 
     #[DataProvider('emptyValueProvider')]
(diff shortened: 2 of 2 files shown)


gsteel:
The work you've done in #125 looks like it'll go some way to addressing some of the problems you mentioned here.

Inheritance in inputs is probably an issue that we'll need to address - we'll also need to look at the `InputInterface` to see if it can be slimmed down making it easier to create implementations.

Not sure that `Input` and derivatives is even useful to users, i.e. most users deal with `InputFilter` and `Input` is more of an implementation detail.

Somewhere, probably in `InputInterface` we need to solidify the order of operations, i.e. in regular inputs, it's filter then validate but file inputs do the reverse - this can be confusing and is IMO non-obvious to users.

My hope with InputFilter is that we can arrive at mostly immutable `InputFilters` and `Input`s at the point of use… It's going to be hard to do this incrementally because the upgrade to SMv4 requires simultaneous upgrade to filter v3 and and validator v3 touching practically _everything_ in the codebase.

Back to the issues with `NotEmpty` validator - The purpose of the behaviour _(of auto adding `NotEmpty` to required inputs)_ is so that validation fails for required inputs that omit any other validators in their specification. I think that we need to find a better way of doing this. The interaction of `required`, `allow_empty`, `continue_if_empty` and "fallback" values needs looking at in greater detail so we can document and explain clear and predictable behaviour for users.

Changing the meaning of `required` could clean this all up quite a lot.

For example, if `required` were to mean - _any_ value was given, empty or not, then a validator is no longer required, because it's a simple case of `array_key_exists('some_key', $input)`... we drop the options `allow_empty` and `continue_if_empty` and simply run the filters and validators when _any_ value is encountered. The responsibility of what is considered valid is then fully moved to the user defined validator chain.

The `required` option would need a companion option of `required_error` or similar to enable customisation of the error message… I'm getting way off track now, suffice to say, there's a lot to discuss.

froschdesign:
> Changing the meaning of `required` could clean this all up quite a lot.
> 
> For example, if `required` were to mean - _any_ value was given, empty or not, then a validator is no longer required, because it's a simple case of `array_key_exists('some_key', $input)`…

This differs from HTML attribute `required`:

> If the element is required and the element's value is the empty string, then the element is suffering from [valueMissing](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState/valueMissing) and the element will match the [:invalid](https://developer.mozilla.org/en-US/docs/Web/CSS/:invalid) pseudo class.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/required#constraint_validation

> …we drop the options `allow_empty` and `continue_if_empty` and simply run the filters and validators when _any_ value is encountered.

But I like this direction.

gsteel:
> This differs from HTML attribute `required`:

But… `laminas-inputfilter` has nothing to do with HTML really - I see its primary use case as filtering and validating arbitrary data structures provided as associative arrays.

It is equally suited to validating JSON payloads and HTML form data, and from a `laminas-form` perspective `isRequired` can still be implemented on inputs to help with rendering markup prior to validation

froschdesign:
> But… `laminas-inputfilter` has nothing to do with HTML really…

Correct, but in combination with laminas-form and it would be good if can adopt the same behaviour and/or we can describe it with the same behaviour like in HTML – _so far the idea_.

ramchale:
> It's going to be hard to do this incrementally

Yeah. I've been exploring this in a very messy branch, and the main blockers I've found are

- `getOption('messageTemplates')` needs changing to `getMessageTemplates()` (which is also deprecated)
- As the code currently stands, `FilterPluginManager` needs injecting in many places, as it’s needed by FilterChain, which is instantiated in Input. The changes you're disucssing above may address this in a way that negates the need to do this
- Classes used in many tests are final and can no longer be mocked. Mostly FilterChain, ValidatorChain, NotEmptyValidator, and in most cases instantiating an real instance works with minor changes

gsteel:
I'm thinking that a possible first step would be to target `FilterChain` and `ValidatorChain` usage across the codebase so that they are not instantiated anywhere without access to the relevant plugin manager, so, without upgrading any deps, we make sure that all filter chains are newed and immediately followed by `->setPluginManager()`. This is potentionally about refactoring the various factories so we can be absolutely sure we are using the same, constructor injected filter & validator plugin containers everywhere.

I'm pretty sure that `InputFilter` itself, which we will want to keep as non-final, will need an explicit hard dependency on either `Factory` or, the plugin managers…

ramchale:
> I'm thinking that a possible first step would be to target `FilterChain` and `ValidatorChain` usage across the codebase so that they are not instantiated anywhere without access to the relevant plugin manager, so, without upgrading any deps, we make sure that all filter chains are newed and immediately followed by `->setPluginManager()`.

Sounds good

I've raised #126 to make a start on adding the necessary plugin managers to Factory. Let me know if this is what you had in mind or if I've got the wrong end of the stick and I'll tidy it up or pivot as appropriate.

ramchale:
I'll close this, as we've moved the discussion elsewhere

PR #126, opened 2025-05-22
Make `InputFilter\Factory` a hard dependency for all Input Filters, Declare `Factory` dependencies and remove various setters/initialisers

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Add plugin managers to Factory for use in Chains

- Add FilterPluginManager and ValidatorPluginManager to Factory for use in setting up other objects
- Add Factory to the constructor of BaseInputFilter
- Add FactoryFactory to instantiate Factory
- Add InputFilterFactory to instantiate InputFilter, CollectionInputFilter, OptionalInputFilter
- Remove InputFilter->setFactory and InputFilter->getFactory
- Remove InputFilterPluginManager->populateFactory


gsteel:
Initial reaction without detailed review - The `InputFilter` is an extension point for users, so keeping the constructor args to a minimum should be a goal. I think it is OK to "hide" the plugin managers in `Factory` and instead require that in the constructor:

```php
class BaseInputFilter {
    public function __construct(
        protected Factory $factory,
    ) {
    }
}
```

When users upgrade, this is going to be a massive pain because currently, InputFilters can be invokable - they are initialised by the inputfilter plugin manager with `setFactory()`… anyhow… `Factory` can be extended to provide new methods such as `createFilterChain(array $config = [])` returning an empty, or populated filter chain with all dependencies satisfied.

We can then focus on `Factory` for creation of `Input`s, `InputFilter`s, `FilterChain`s and `ValidatorChain`s, slowly ridding these components of setter injection and killing off the "initialisers" in `InputFilterPluginManager`.

Given, the factory is now always available inside InputFilter, inheritors using `init()` will be mostly unaffected, apart from the problems associated with changes to the constructor. i.e. `$this->add(['spec'], 'name')` from an input filter context will still just proxy to `Factory::createInput|createInputFilter()`

src/Factory.php
```diff
@@ -37,13 +39,16 @@ class Factory
     /** @var ValidatorChain|null */
     protected $defaultValidatorChain;
 
-    /** @var InputFilterPluginManager|null */
-    protected $inputFilterManager;
+    public function __construct(
+        protected ?FilterPluginManager $filterPluginManager,
+        protected ?ValidatorPluginManager $validatorPluginManager,
+        protected ?InputFilterPluginManager $inputFilterManager = null
```
gsteel: I think that `Factory` should be `final` with private non-nullable props. Potentially it could also be `@internal` as an implementation detail of `InputFilter::add()`, but that might not be possible.

src/Factory.php
```diff
@@ -128,7 +133,11 @@ public function setInputFilterManager(InputFilterPluginManager $inputFilterManag
     public function getInputFilterManager()
     {
         if (null === $this->inputFilterManager) {
-            $this->inputFilterManager = new InputFilterPluginManager(new ServiceManager());
+            $sm = new ServiceManager();
+            // TODO - is there a better way of getting this into the InputFilter constructor
+            $sm->setService(self::class, $this);
+
+            $this->inputFilterManager = new InputFilterPluginManager($sm);
```
gsteel: I'd just bin the getter and setter for the plugin manager as per `__construct` comment.

src/InputFilterAbstractServiceFactory.php
```diff
@@ -94,7 +94,10 @@ protected function getInputFilterFactory(ContainerInterface $container)
             return $this->factory;
         }
 
-        $this->factory  = new Factory();
+        $this->factory  = new Factory(
+            $this->getFilterPluginManager($container),
+            $this->getValidatorPluginManager($container)
+        );
```
gsteel: With `Factory` having a constructor dependency on the plugin manager, it should get its own DI Factory, i.e. `Laminas\InputFilter\Factory\FactoryFactory` (😂), so most of the methods here can be replaced with `$container->get('whatever')`

src/InputFilterFactory.php
```diff
@@ -0,0 +1,38 @@
...
+    public function __invoke(ContainerInterface $container, $requestedName, ?array $options = null)
+    {
+        $factory = $container->get(Factory::class);
+
+        return null === $options ?
+            new $requestedName($factory) :
+            new $requestedName($factory, $options);
+    }
+
+    private function getFilterPluginManager(ContainerInterface $container): FilterPluginManager
```
gsteel: I think it should crash rather than `new FilterPlugins` - We don't own `filter` or `validator` here, so we should just assume they are in the container, or crash accordingly, i.e leave `ServiceNotFoundException` unchecked

src/InputFilterFactory.php
```diff
@@ -0,0 +1,38 @@
...
+use Psr\Container\ContainerInterface;
+
+class InputFilterFactory implements FactoryInterface
+{
+    public function __invoke(ContainerInterface $container, $requestedName, ?array $options = null)
+    {
+        $factory = $container->get(Factory::class);
+
+        return null === $options ?
+            new $requestedName($factory) :
```
gsteel: Psalm is correctly complaining that it has no idea what `$requestedName` is.

If we limit the possibilities via a constant such as

```php
private const CAN_CREATE = [
    InputFilter::class => InputFilter::class,
    Collection::class => Collection::class,
    // etc…
];

// and later on…

$class = self::CAN_CREATE[$requestedName] ?? null;
// if not null
return new $class($container->get(Factory::class));
```

Not sure if `$options` is neccessary.

This factory will not be useful to inheritors of `InputFilter` because `__constructor` signatures deviate from parent classes freely. For that, we might ship a `ReflectionBasedAbstractFactory` specifically for InputFilter inheritors so users don't have to create factories all the time.

src/InputFilterFactory.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Laminas\InputFilter;
+
+use Laminas\Filter\FilterPluginManager;
+use Laminas\ServiceManager\Factory\FactoryInterface;
+use Laminas\Validator\ValidatorPluginManager;
+use Psr\Container\ContainerInterface;
+
+class InputFilterFactory implements FactoryInterface
```
gsteel: `final` and `@internal`

FYI, you can add

```php
/**
 * @psalm-internal Laminas\InputFilter
 * @psalm-internal LaminasTest\InputFilter
 */
```

To prevent Psalm complaining from different contexts. Adding just `@internal` causes issues in test code.

src/InputFilterPluginManager.php
```diff
@@ -12,6 +12,7 @@
 use Laminas\ServiceManager\ServiceManager;
 use Laminas\Stdlib\InitializableInterface;
 use Laminas\Validator\ValidatorPluginManager;
+use phpDocumentor\Reflection\Types\This;
```
gsteel: Stray import here

gsteel (changes requested):
Thanks @ramchale - Overall I think this is a solid first step. If we can get the various issues sorted and the tests passing, we should be in good shape to get SMv4 support working 👍 

Making `Factory` `final` with required props might need some new test infrastructure for more convenience around the test suite.

ramchale:
Thanks for the feedback so far @gsteel 
I've not had as much time as I'd like to look at this over the last couple of weeks, but I don't think this is too far off now. ~~Few failing tests that I think are all the same underlying issue, and~~ I need to do a tidying pass after that.
I've introduced a test helper for instantiating Factory for tests, but if there's an approach that's more consistent with other Laminas libraries let me know.

src/FactoryFactory.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+namespace Laminas\InputFilter;
+
+use Laminas\Filter\FilterPluginManager;
+use Laminas\ServiceManager\Factory\FactoryInterface;
+use Laminas\Validator\ValidatorPluginManager;
+use Psr\Container\ContainerInterface;
+
+final class FactoryFactory implements FactoryInterface
```
gsteel: Because this is a regular invokable factory, we can drop the factory interface and the unnecessary method args.

Where possible, make all factories internal with `@psalm-internal` for both `tests` and `src` - this is so that we don't have to bother with BC for basic wiring.

src/InputFilter.php
```diff
@@ -15,32 +15,14 @@
...
-     * @return InputFilter
-     */
-    public function setFactory(Factory $factory)
-    {
-        $this->factory = $factory;
-        return $this;
-    }
-
     /**
+     * TODO - Deprecate and remove?
```
gsteel: Yes!

src/InputFilterFactory.php
```diff
@@ -0,0 +1,36 @@
...
+use Psr\Container\ContainerInterface;
+use RuntimeException;
+
+use function sprintf;
+
+/**
+ * @psalm-internal Laminas\InputFilter
+ * @psalm-internal LaminasTest\InputFilter
+ */
+final class InputFilterFactory implements FactoryInterface
```
gsteel: It might be better to implement `AbstractFactoryInterface` here with its `canCreate` method - that way, excpetions thrown will be consistent IIRC - I might be wrong, in which case this is fine.

src/InputFilterPluginManager.php
```diff
@@ -113,8 +113,8 @@ public function populateFactory($containerOrInputFilter, $inputFilter = null)
             return;
         }
 
-        $factory = $inputFilter->getFactory();
-        $factory->setInputFilterManager($this);
+//        $factory = $inputFilter->getFactory();
+//        $factory->setInputFilterManager($this);
```
gsteel: Let's pre-emptively delete the initializer entirely

gsteel (changes requested):

gsteel:
> I've introduced a test helper for instantiating Factory for tests, but if there's an approach that's more consistent with other Laminas libraries let me know.

This is fine 👍 - I might have called it something more generic like `ContainerHelper` and had a static method to generate a `ServiceManager` instance using `ConfigProvider::__invoke()` with `config` service already defined like this: https://github.com/laminas/laminas-view/blob/25c6097c09b7d1702c5e2362d5a3f99a00cb7f78/test/GenerateServiceManager.php

What we've got here is fine though, so no need to change it.

ramchale:
Thanks again @gsteel . I've addressed the feedback so far and squashed this to remove the WIP commits

src/InputFilterAbstractServiceFactory.php
```diff
@@ -76,59 +73,32 @@ public function canCreateServiceWithName(ServiceLocatorInterface $container, $na
...
         $validatorChain = $this->factory->getDefaultValidatorChain();
         assert($filterChain instanceof FilterChain);
         assert($validatorChain instanceof ValidatorChain);
 
-        $filterChain->setPluginManager($this->getFilterPluginManager($container));
-        $validatorChain->setPluginManager($this->getValidatorPluginManager($container));
-
-        $this->factory->setInputFilterManager($container->get(InputFilterPluginManager::class));
+        $filterChain->setPluginManager($container->get(FilterPluginManager::class));
+        $validatorChain->setPluginManager($container->get(ValidatorPluginManager::class));
```
gsteel: All of this is done in `Factory::__construct()` right? so all of this can get in the bin, the entire method could go in the bin along with the property - just fetch Factory from the container where it's needed

src/InputFilterPluginManager.php
```diff
@@ -88,35 +86,6 @@ class InputFilterPluginManager extends AbstractPluginManager
...
-        $inputFilter = $inputFilter ?: $containerOrInputFilter;
-
-        if (! $inputFilter instanceof InputFilter) {
-            return;
-        }
-
-        $factory = $inputFilter->getFactory();
-        $factory->setInputFilterManager($this);
-    }
-
```
gsteel, line 119: Can `populateFactoryPluginManagers()` and all its usages go in the bin too - because `Factory::__construct()` already does the work here

gsteel (changes requested):
Thanks @ramchale - LGTM other than some now redundant code that initialises default filter|validator chains

src/InputFilterAbstractServiceFactory.php
ramchale: Cool. Make sense. Thanks

src/Factory.php
```diff
@@ -32,11 +32,8 @@
  */
 final class Factory
 {
-    /** @var FilterChain|null */
-    protected $defaultFilterChain;
-
-    /** @var ValidatorChain|null */
-    protected $defaultValidatorChain;
+    protected ?FilterChain $defaultFilterChain;
+    protected ?ValidatorChain $defaultValidatorChain;
```
gsteel, line 36: There's no point in `protected` in a `final` class, also, these props cannot possibly be null, unless there's a `setWhateverChain(Type|null $chain)` method (?)

gsteel (changes requested):

src/Factory.php
ramchale: Currently there's `clearDefaultFilterChain` and `clearDefaultValidatorChain` methods that set these to null. It would probably be more intuitive to allow `setDefaultFilterChain` and `setDefaultValidatorChain` to accept null and remove the clear methods.

gsteel: No worries, we can clean up that stuff in another patch 👍

gsteel (approved):
Thanks @ramchale - You can merge this now ;)

PR #127, opened 2025-05-23
WIP - SM4 compatibility, type declarations, and other refactors

adambalint-srg:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

It's only a draft PR. For example tests are not refactored, and not tested. I did the refactor of the package to be compatible with SM4, and updated `laminas-filter` and `laminas-validator` packages.

The most of this PR is the refactored type declarations, and removing of some redundant doc block. 

The factories and plugin managers have been refactored to be compatible with SM4

The `Factory` class must be refactored because of changes in `laminas-filter` and `laminas-validator` package. It affects the input validator and filter chains, so `Factory` now coming from the Container.

This PR is only a question now, if it is a good way, and I can continue this contribution, or I should forget these big refactors :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,18 +32,18 @@
     },
     "require": {
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
-        "laminas/laminas-filter": "^2.19",
-        "laminas/laminas-servicemanager": "^3.21.0",
-        "laminas/laminas-stdlib": "^3.19",
-        "laminas/laminas-validator": "^2.60.0"
+        "laminas/laminas-filter": "^3.0.0",
+        "laminas/laminas-servicemanager": "^4.4",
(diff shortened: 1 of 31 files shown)


arhimede:
@gsteel if you can take a look at this one, this library is the last major one left to be upgraded to SM4

gsteel:
Appreciate the effort here @adambalint-srg, but going for the SMv4 upgrade as a first step is just too big TBH. There's a lot of room for improvement and refactoring oportunities in this lib and resolving some of those upfront should make the transition to SMv4 validator 3 and filter 3 a lot smoother and cleaner.

We also need to work on migration guides with detailed BC breaks too.

Take a look at #126 as a first step - comments welcome.

When time allows, I will try to work up a road map / todo list for this component 👍

PR #128, opened 2025-05-26
Add Makefile and Improved QA Tooling

gsteel:
- Declare requirement of PSR\Container in dependencies
- Add CI Job for composer dependencies

This will be particularly useful during work on 3.x

`make qa` also validates docs with markdownlint along with all the other QA checks.

As a dependency change, we can release this in 2.33.0 and get it merged up into 3.0.x

Please don't release just yet though - I've got some prep on the docs tree to do.


Makefile
```diff
@@ -0,0 +1,81 @@
+# Run `make` (no arguments) to get a short description of what is available
+# within this `Makefile`.
+
+MKDOCS_IMAGE_ID := $(shell docker images -q laminas/mkdocs | xargs)
+MDLINT_FILE = https://raw.githubusercontent.com/laminas/laminas-continuous-integration-action/refs/heads/1.43.x/setup/markdownlint/markdownlint.json
+
+help: ## shows this help
+	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_\-\.]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
+.PHONY: help
```
Ocramius, line 9: Isn't there a wildcard target for phony targets?

Ocramius (approved):

Makefile
gsteel: I believe you can put all the "phonies" on one line, i.e. `.PHONY: foo bar baz` - is that what you mean?

Ocramius: Something like `.PHONY: *` at the very top?

gsteel: Ahh OK. That makes sense. In this case, there are a few non-phony targets that fail with that. Any idea how to negate those?

Ocramius: Sorry, unaware of that :D 

I only ever **abuse** `Makefile` :D

PR #133, opened 2025-06-28
Loosen coupling of `CollectionInputFilter` with `NotEmpty` validator

ramchale:
- Make getNotEmptyValidator private and add public setIsRequiredValidationMessage method
- Remove deprecated getOption method call in prepareRequiredValidationFailureMessage

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | yes
| QA            | yes

### Description

The call to `$notEmptyValidator->getOption('messageTemplates')` in CollectionInputFilter will need replacing to enable moving to ServiceManager v4. We could replace this with a call to `getMessageTemplates` but this is also deprecated. A direct call to the validator with an invalid value should achive the same results without CollectionInputFilter needing to understand the internals of the NotEmpty validator.

### DIFF
--- a/src/CollectionInputFilter.php
+++ b/src/CollectionInputFilter.php
@@ -37,9 +37,7 @@ class CollectionInputFilter extends InputFilter
 
     /** @var BaseInputFilter|null */
     protected $inputFilter;
-
-    /** @var NotEmpty|null */
-    protected $notEmptyValidator;
+    private ?string $emptyErrorMessage = null;
 
     /**
      * Set the input filter to use when looping the data
@@ -85,13 +83,23 @@ class CollectionInputFilter extends InputFilter
 
     /**
      * Set if the collection can be empty
-     *
-     * @param bool $isRequired
-     * @return $this
      */
-    public function setIsRequired($isRequired)
+    public function setIsRequired(bool $isRequired): static
     {
         $this->isRequired = $isRequired;
+
+        return $this;
+    }
+
+    /**
+     * Set a custom error message for the collection being empty.
+     * If not called, CollectionInputFilter will default to the NotEmpty validators IS_EMPTY message
+     *
+     * @param non-empty-string $message
+     */
+    public function setIsRequiredValidationMessage(string $message): static
+    {
+        $this->emptyErrorMessage = $message;
 
         return $this;
     }
@@ -171,38 +179,6 @@ class CollectionInputFilter extends InputFilter
     }
 
     /**
-     * Retrieve the NotEmpty validator to use for failed "required" validations.
-     *
-     * This validator will be used to produce a validation failure message in
-     * cases where the collection is empty but required.
-     *
-     * @return NotEmpty
-     */
-    public function getNotEmptyValidator()
-    {
-        if ($this->notEmptyValidator === null) {
-            $this->notEmptyValidator = new NotEmpty();
-        }
-
-        return $this->notEmptyValidator;
-    }
-
-    /**
-     * Set the NotEmpty validator to use for failed "required" validations.
-     *
-     * This validator will be used to produce a validation failure message in
-     * cases where the collection is empty but required.
-     *
-     * @return $this
-     */
-    public function setNotEmptyValidator(NotEmpty $notEmptyValidator)
-    {
-        $this->notEmptyValidator = $notEmptyValidator;
-
-        return $this;
-    }
-
-    /**
      * @inheritDoc
      */
     public function isValid($context = null)
@@ -212,7 +188,7 @@ class CollectionInputFilter extends InputFilter
         $valid                    = true;
 
         if ($this->getCount() < 1 && $this->isRequired) {
-            $this->collectionMessages[] = $this->prepareRequiredValidationFailureMessage();
+            $this->collectionMessages[] = $this->getEmptyValidationErrorMessages();
             $valid                      = false;
         }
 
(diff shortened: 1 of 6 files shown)


ramchale:
Just to check. Can `setNotEmptyValidator` and `getNotEmptyValidator' be removed from the public interface of `CollectionInputFilter` in v3, or is there use cases that I'm unaware of that require these to be public?

alexmerlin:
> Just to check. Can `setNotEmptyValidator` and `getNotEmptyValidator' be removed from the public interface of `CollectionInputFilter` in v3, or is there use cases that I'm unaware of that require these to be public?

@ramchale IMO they can stay private inside of `CollectionInputFilter`.
I'm not aware of any use case in related packages (checked in `laminas-form`, `laminas-input` and `laminas-validator`).
Maybe @gsteel can confirm?

src/CollectionInputFilter.php
```diff
@@ -85,17 +85,25 @@ public function getInputFilter()
...
     {
         $this->isRequired = $isRequired;
 
         return $this;
     }
 
+    /**
+     * Set a custom error message for the collection being empty.
+     * If not called, CollectionInputFilter will default to the NotEmpty validators IS_EMPTY message
+     */
```
gsteel, line 99:
```suggestion
     * @param non-empty-string $message
     */
```

src/CollectionInputFilter.php
```diff
@@ -343,15 +334,7 @@ public function getUnknown()
...
-        $message    = $templates[NotEmpty::IS_EMPTY];
-        $translator = $notEmptyValidator->getTranslator();
-
-        return [
-            NotEmpty::IS_EMPTY => $translator
-                ? $translator->translate($message, $notEmptyValidator->getTranslatorTextDomain())
-                : $message,
-        ];
+        $notEmptyValidator(null);
+        return $notEmptyValidator->getMessages();
```
gsteel: The problem with this is now it is impossible to translate the error message, because it is not even possible to fetch the `NotEmpty` instance in order to set the validator _(I don't think that is possible in v3 anyhow)_.

Might it be better to, in `getNotEmptyValidator()` call:

```php
$this->factory->getValidatorPluginManager()->build(
    NotEmpty::class,
   [
       'messages' => [
           NotEmpty::IS_EMPTY => $this->requiredValidationMessage,
       ],
   ],
);

```

…instead of `new NotEmpty()`.

Doing this will mean that if the user wishes to use a translator, via some kind of validator initialiser, that translator will be used when necessary.

Edit… Thinking about this some more…

The validator is not actually executed with the input data at any point, so its purpose is purely to extract a single error message.

I'd suggest:

- removing the `getNotEmptyValidator` method, along with `$this->notEmptyValidator` property.
- renaming this method to `private function getEmptyValidationErrorMessages(): array`
- Storing the configured "empty" error message in a property such as `private string|null $emptyErrorMessage = null`
- and changing the body of this method to something like:

```php
/** @return array<string, string> */
private function getEmptyValidationErrorMessages(): array
{
    $options = $this->emptyErrorMessage === null
        ? []
        : ['messages' => [NotEmpty::IS_EMPTY => $this->emptyErrorMessage]];

    $validator = $this->factory->getValidatorPluginManager()->build(NotEmpty::class, $options);
    
    $validator->isValid(null);
    
    return $validator->getMessages();
}
```

src/Input.php
```diff
@@ -463,8 +463,9 @@ protected function prepareRequiredValidationFailureMessage()
         }
 
         /** @psalm-var array<string, string> $templates */
-        $templates  = $notEmpty->getOption('messageTemplates');
-        $message    = $templates[NotEmpty::IS_EMPTY];
+        $templates = $notEmpty->getOption('messageTemplates');
+        $message   = $templates[NotEmpty::IS_EMPTY];
+        /** @psalm-suppress DeprecatedInterface */
```
gsteel, line 468: The entire body of this method needs to be reconsidered.

```
$validator = $chain->hasANotEmptyValidator()
    ? $chain->getTheNotEmptyValidator()
    : $this->validators->makeADefaultNotEmtpy();

$validator->isValid(null);

return $validator->getMessages();
```

Then we can drop all the mucking about with translators and internal options inside `NotEmpty` that we have no business mucking about with.

test/CollectionInputFilterTest.php
```diff
@@ -746,32 +747,10 @@ public function testDuplicatedErrorMessages(): void
...
-
-        self::assertFalse($this->inputFilter->isValid());
-
-        self::assertEquals([
-            [NotEmpty::IS_EMPTY => $message],
-        ], $this->inputFilter->getMessages());
+        self::assertInstanceOf(
+            NotEmpty::class,
+            (new ReflectionMethod($this->inputFilter, 'getNotEmptyValidator'))->invoke($this->inputFilter)
+        );
```
gsteel: I think we need to test here that a collection built via factory with…

`$factory->createInputFilter($configToMakeACollection)`

will yield the expected error message, when…

`$configToMakeACollection = ['required_message' => 'Whatever', ...];`

_and_

We get a non-empty error message when no custom error message has been set.

_and_

The setter for `setEmptyMessage()` or whatever the method is, affects the error message appropriately as per factory-backed construction.

We should avoid referring to any part of the NotEmpty validator in tests here.

gsteel (changes requested):
Thanks @ramchale - Sorry it took me a while to get to reviewing this!

src/Input.php
ramchale: 👍 I can pick this up as a seperate change

src/CollectionInputFilter.php
ramchale: Thanks

src/CollectionInputFilter.php
ramchale: Cool. Sounds sensible

ramchale:
@gsteel thanks for the suggestions. Let me know if I've missed anything.
If you think this is ready, it might be worth merging 2.34.x into 3.0.x and rebasing and squashing this PR to keep the history slightly cleaner

gsteel (approved):
Thanks @ramchale LGTM 👍 

I agree about a merge up of 2.34.x - I'll do that now so you can rebase here when you're ready

ramchale:
@gsteel thanks. That's rebased and ready for re-review

gsteel (approved):

PR #135, opened 2025-07-08
Replace mocked FilterChain with instance in unit tests

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation |no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | yes
| QA            | yes

### Description

FilterChain is final and from laminas-filter v3 onwards this is enforced preventing mocking in unit tests.

This replaces the mocks with instances, and reduces some test setup duplication

### DIFF
--- a/test/TestHelper.php
+++ b/test/TestHelper.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace LaminasTest\InputFilter;
 
+use Laminas\Filter\FilterChain;
 use Laminas\Filter\FilterPluginManager;
 use Laminas\InputFilter\Factory;
 use Laminas\InputFilter\InputFilterPluginManager;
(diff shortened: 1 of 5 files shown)


test/TestHelper.php
```diff
@@ -46,4 +45,29 @@ public static function createValidatorChain(mixed $value, bool $isValid): Valida
     {
         return (new ValidatorChain())->attach(self::createValidatorMock($isValid, $value));
     }
+
+    public static function createFilterChain(mixed $originalValue, mixed $filteredValue): FilterChain
```
gsteel: Given that `FilterChain` will have ctor deps in 3.0, I'd suggest keeping `createFilterChain` a method with no arguments, and rename this method to something like `createFilterChainFixture()`, you could then replace all `new FilterChain` calls with `TestHelper::createFilterChain()` pre-emptively.

gsteel (approved):
Thanks @ramchale - Just a suggestion as this can always be changed in another patch. I'll leave it to you to amend or merge 👍

test/TestHelper.php
ramchale: Thanks. I was in two minds about that, but if two of us are considering it it's probably the right approach 😄

PR #125, opened 2025-05-17
Remove skipped input tests

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Remove several skipped tests, and remove inheritance from InputTest.

Although this adds more duplication, the ability to independently make changes to these tests without unintended side effects should make future maintenance easier and simplifies some of the test logic.

PR #136, opened 2025-07-12
Replace mocked FilterChain with instance in InputTest and ArrayInputTest

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | yes
| QA            | yes

### Description

NotEmpty is final and from laminas-validator v3 onwards this is enforced preventing mocking in unit tests.

This replaces the mocks with instances in InputTest and ArrayInputTest, and corrects `new Input` to `new ArrayInput` in some tests

### DIFF
--- a/test/ArrayInputTest.php
+++ b/test/ArrayInputTest.php
@@ -29,8 +29,6 @@ use function array_diff_key;
 use function array_merge;
 use function array_pop;
 use function count;
-use function current;
-use function is_array;
 use function iterator_to_array;
 use function json_encode;
 use function sprintf;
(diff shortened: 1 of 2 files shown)


test/ArrayInputTest.php
```diff
@@ -482,18 +480,14 @@ public function testRequiredWithoutFallbackAndValueNotSetProvidesNotEmptyValidat
 
     public function testRequiredWithoutFallbackAndValueNotSetProvidesAttachedNotEmptyValidatorIsEmptyErrorMessage(): void // phpcs:ignore
     {
-        $input = new Input();
+        $input = new ArrayInput();
```
gsteel, line 483: 😅 Was this copy/paste from a recent patch or did this test case always use the wrong input type?

gsteel (approved):
Thanks @ramchale 👍

test/ArrayInputTest.php
ramchale: Mostly the copy paste. Some of the tests were previously inherited from InputTest until #125 , so ArrayInput wasn't covered as they don't use the instance from setup, then incorrectly covered due to copy/paste error, now finally actually covered 😓

ramchale: This corrects the same issue in the most recent x 2 brannch #137

PR #137, opened 2025-07-14
Correct ArrayInputTest instantiating test class for special cases

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

This fixes several cases that do not use the instance created by the setUp method and were instantiating Input rather than ArrayInput


gsteel (approved):
Thanks @ramchale - That'd gone unnoticed for a long time!!

PR #142, opened 2025-07-22
Simplify Input->prepareRequiredValidationFailureMessage

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | yes
| QA            | yes

### Description

In Input->prepareRequiredValidationFailureMessage, use the logic in NotEmpty to build the error message and translate it where necessary, rather than extracting the message templates and translator from NotEmpty.

This makes prepareRequiredValidationFailureMessage less brittle to chages to NotEmpty internals, and removes deprecations for laminas-validator v3

### DIFF
--- a/src/Input.php
+++ b/src/Input.php
@@ -7,7 +7,6 @@ namespace Laminas\InputFilter;
 use Laminas\Filter\FilterChain;
 use Laminas\ServiceManager\AbstractPluginManager;
 use Laminas\Validator\NotEmpty;
-use Laminas\Validator\Translator\TranslatorInterface;
 use Laminas\Validator\ValidatorChain;
 
 use function class_exists;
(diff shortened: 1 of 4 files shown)


ramchale:
@gsteel not sure this quite covers what you suggested in https://github.com/laminas/laminas-inputfilter/pull/133#discussion_r2192831310 so let me know if there's a wider change that would be preferable

src/Input.php
```diff
@@ -462,18 +461,10 @@ protected function prepareRequiredValidationFailureMessage()
             }
         }
 
-        /** @psalm-var array<string, string> $templates */
-        $templates = $notEmpty->getOption('messageTemplates');
-        $message   = $templates[NotEmpty::IS_EMPTY];
-        /** @psalm-suppress DeprecatedInterface */
-        $translator = $notEmpty->getTranslator();
+        $validator = $notEmpty ?: $chain->plugin(NotEmpty::class);
```
gsteel, line 464: `$chain->plugin()` should be gone in validator 3.x

gsteel (approved):
Thanks @ramchale LGTM 👍

src/Input.php
ramchale: Ah okay. I'll take a look at that as well

PR #143, opened 2025-07-23
Remove FileInput inheritance from Psr and HttpServer implementations

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | yes
| QA            | yes

### Description

The inheritance was only used to allow access to protected properties of FileInput. This refactor removes the need for this and makes the flow of logic slightly easier to follow.


ramchale:
The intent is to make this a little easier to understand and maintain going forward. If people are happy with the approach then the tests in PsrFileInputDecoratorTest and HttpServerFileInputDecoratorTest should mostly move into a FileInputTest file and be de-duplicated as much as possible. We could also consider renaming PsrFileInputDecorator and HttpServerFileInputDecorator to be clearer on their purpose

src/FileInput/HttpServerFileInputDecorator.php
```diff
@@ -29,15 +29,10 @@
  *
  * @final
  */
-class HttpServerFileInputDecorator extends FileInput implements FileInputDecoratorInterface
+class HttpServerFileInputDecorator implements FileInputDecoratorInterface
```
gsteel: Can we use the `final` keyword and make this `@internal`?

FYI, instead of `@internal` you can

```php
/**
 * @psalm-internal Laminas\InputFilter
 * @psalm-internal LaminasTest\InputFilter
 */
```

… to minimise Psalm issues

src/FileInput/FileInputDecoratorInterface.php
```diff
@@ -13,22 +16,11 @@
  */
 interface FileInputDecoratorInterface
```
gsteel: `@internal`?

src/FileInput.php
```diff
@@ -25,13 +25,13 @@
...
-class FileInput extends Input
+final class FileInput extends Input
 {
     /** @var bool */
-    protected $isValid = false;
+    private $isValid = false;
 
     /** @var bool */
-    protected $autoPrependUploadValidator = true;
+    private $autoPrependUploadValidator = true;
```
gsteel: Can this property be made immutable rather than passing it to the decorator constructor by ref? By ref stuff really sucks and is horrible to test - ideally we could drop the getter and setter, or, make the input itself the constructor arg so the decorator can call the getter when it matters.

That said, you're passing the validator chain as a method arg - why not mutate the validator chain inside `FileInput` prior to executing the decorator when necessary, then the decorator doesn't need access to this property at all - i.e. the decorator accepts the validator chain as-is.

gsteel (commented):
This will be a huge improvement in readability and maintainability 👍 

I'm in favour of renaming the 'Decorators' and explicitly marking all of that stuff as internal implementation detail.

src/FileInput.php
ramchale: ~~If we remove get/setAutoPrependUploadValidator from FileInput then yep, I can handle autoPrependUploadValidator in a less horrible way 😄~~

~~The initial approach I looked at ran into issues with calls to FileInput->setValidatorChain after construction of the decorator. The injectUploadValidator methods in each decorator are putting in different validators, but I'll give it more of a look, as there's probably a nicer way of handling this~~

Scratch all that, clearly needed to walk away from the laptop for a bit. Given 'fileuploadfile' is an alias for the UploadFile validator this should work fine

ramchale:
@gsteel I think it ready for another look.
Despite raising it, I can't actually think of a better name for the decorators, they're handling some validation and filtering so it'd need to be something that neatly encompasses both.

test/FileInput/FileInputTest.php
```diff
@@ -0,0 +1,1146 @@
...
+
+    #[DataProvider('validSingleValueProvider')]
+    public function testRequiredUploadValidatorValidatorNotAddedWhenOneExists(mixed $raw): void
+    {
+        $this->input->setAutoPrependUploadValidator(true);
+        self::assertTrue($this->input->getAutoPrependUploadValidator());
+        self::assertTrue($this->input->isRequired());
+        $this->input->setValue($raw);
+
+        $uploadMock = $this->createMock(Validator\File\UploadFile::class);
```
gsteel: Pretty sure this is final in 3.0 so mocking is out of the window. Can you compose a chain without a mock here, or, fix it when we upgrade to validator 3.x.

gsteel (approved):
Thanks @ramchale - Just one note about mocking the upload validator, but otherwise LGTM 👍

gsteel:
> Despite raising it, I can't actually think of a better name for the decorators, they're handling some validation and filtering so it'd need to be something that neatly encompasses both.

Maybe go with something generic such as `Helper`. I think it's a good idea to change the name because they really aren't decorators at all.

test/FileInput/FileInputTest.php
ramchale: Good point. I've switched it to use values that fail validation to avoid having to fake uploading a file, but it is now using an actual instance of UploadFile

ramchale:
> > Despite raising it, I can't actually think of a better name for the decorators, they're handling some validation and filtering so it'd need to be something that neatly encompasses both.
> 
> Maybe go with something generic such as `Helper`. I think it's a good idea to change the name because they really aren't decorators at all.

👍  I've gone with Handler

ramchale:
I've squashed this, so if you're happy with the last couple of bits, I think it's good to merge

src/FileInput/FileInputDecoratorInterface.php
```diff
@@ -4,31 +4,26 @@
...
  * FileInputInterface defines expected methods for filtering uploaded files.
  *
  * FileInput will consume instances of this interface when filtering files,
  * allowing it to switch between SAPI uploads and PSR-7 UploadedFileInterface
  * instances.
+ *
+ * @psalm-internal Laminas\InputFilter
+ * @psalm-internal LaminasTest\InputFilter
  */
 interface FileInputDecoratorInterface
```
gsteel: You should also rename this to something like `FileInputHandlerInterface`

src/FileInput.php
```diff
@@ -79,7 +77,7 @@ public function getValue()
         if ($this->implementation === null) {
             return $this->value;
         }
-        return $this->implementation->getValue();
+        return $this->implementation->filterValue($this->value, $this->isValid, $this->getFilterChain());
```
gsteel: Can we rename this private prop `$handler` instead of `$implementation`?

src/FileInput.php
```diff
@@ -172,20 +175,42 @@ private function createDecoratorImplementation($value)
     {
```
gsteel, line 171: And this method should be renamed too.

src/FileInput/FileInputDecoratorInterface.php
```diff
@@ -4,31 +4,26 @@
...
 {
-    /**
-     * Checks if the raw input value is an empty file input eg: no file was uploaded
-     *
-     * @param mixed $rawValue
-     * @return bool
-     */
-    public static function isEmptyFileDecorator($rawValue);
+    /** Checks if the raw input value is an empty file input eg: no file was uploaded */
+    public static function isEmptyFileDecorator(mixed $rawValue): bool;
```
gsteel: Can this just be `isEmptyFile`

gsteel (changes requested):

ramchale:
@gsteel thanks. Not sure how I missed those 😓 , should be cleaned up now

gsteel (approved):
Thanks @ramchale 👍

PR #144, opened 2025-07-27
Add FilterChain and ValidatorChain to the Input constructor

ramchale:
Raised as draft as #143 will need to be merged first to avoid conflict

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | yes
| QA            | yes

### Description

This is required in laminas/laminas-filter v3 to construct the default FilterChain

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -7,15 +7,18 @@ namespace Laminas\InputFilter;
 use Laminas\Filter\FilterChain;
 use Laminas\Filter\FilterInterface;
 use Laminas\Filter\FilterPluginManager;
+use Laminas\ServiceManager\ServiceManager;
 use Laminas\Stdlib\ArrayUtils;
 use Laminas\Validator\ValidatorChain;
 use Laminas\Validator\ValidatorInterface;
 use Laminas\Validator\ValidatorPluginManager;
+use Psr\Container\ContainerInterface;
 use Traversable;
 
 use function assert;
 use function class_exists;
 use function get_debug_type;
+use function is_a;
 use function is_array;
 use function is_callable;
 use function is_int;
@@ -35,8 +38,35 @@ final class Factory
     protected ?FilterChain $defaultFilterChain;
     protected ?ValidatorChain $defaultValidatorChain;
 
+    public static function new(ContainerInterface|null $container = null): self
+    {
+        $container = $container ?? new ServiceManager();
+
+        if ($container->has(self::class)) {
+            return $container->get(self::class);
+        }
+
+        $factory = new Factory(
+            $container->has(FilterPluginManager::class)
+            ? $container->get(FilterPluginManager::class)
+            : new FilterPluginManager($container),
+            $container->has(ValidatorPluginManager::class)
+            ? $container->get(ValidatorPluginManager::class)
+            : new ValidatorPluginManager($container),
+            $container->has(InputFilterPluginManager::class)
+            ? $container->get(InputFilterPluginManager::class)
+            : new InputFilterPluginManager($container),
+        );
+
+        if ($container instanceof ServiceManager) {
+            $container->setService(self::class, $factory);
+        }
+
+        return $factory;
+    }
+
     public function __construct(
-        FilterPluginManager $filterPluginManager,
+        private readonly FilterPluginManager $filterPluginManager,
         private readonly ValidatorPluginManager $validatorPluginManager,
         private readonly InputFilterPluginManager $inputFilterPluginManager
     ) {
@@ -154,7 +184,19 @@ final class Factory
             ));
         }
 
-        $input = $managerInstance ?: new $class();
+        if (is_a($class, Input::class, true)) {
+            $filterChain = new FilterChain();
+            /** @psalm-suppress DeprecatedMethod removal will be done in Service Manager 4 upgrade */
+            $filterChain->setPluginManager($this->filterPluginManager);
+
+            $validatorChain = new ValidatorChain();
+            $validatorChain->setPluginManager($this->validatorPluginManager);
+
+            /** @psalm-suppress UnsafeInstantiation */
+            $input = $managerInstance ?: new $class($filterChain, $validatorChain);
+        } else {
+            $input = $managerInstance ?: new $class();
+        }
 
         if ($input instanceof InputFilterInterface) {
             return $this->createInputFilter($inputSpecification);
--- a/src/Input.php
+++ b/src/Input.php
@@ -28,17 +28,11 @@ class Input implements
     /** @var string|null */
     protected $errorMessage;
 
-    /** @var null|FilterChain */
-    protected $filterChain;
-
     /** @var bool */
     protected $notEmptyValidator = false;
 
     /** @var bool */
     protected $required = true;
-
-    /** @var null|ValidatorChain */
-    protected $validatorChain;
 
     /** @var mixed */
     protected $value;
@@ -56,9 +50,11 @@ class Input implements
     /** @var bool */
     protected $hasFallback = false;
 
-    /** @param null|string $name */
-    public function __construct(protected $name = null)
-    {
+    public function __construct(
+        protected FilterChain $filterChain,
+        protected ValidatorChain $validatorChain,
+        protected ?string $name = null
+    ) {
     }
 
     /**
@@ -218,14 +214,8 @@ class Input implements
         return $this->errorMessage;
     }
 
-    /**
-     * @return FilterChain
-     */
-    public function getFilterChain()
+    public function getFilterChain(): FilterChain
     {
-        if (! $this->filterChain) {
-            $this->filterChain = new FilterChain();
-        }
         return $this->filterChain;
     }
 
@@ -258,9 +248,6 @@ class Input implements
      */
     public function getValidatorChain()
     {
-        if (! $this->validatorChain) {
-            $this->validatorChain = new ValidatorChain();
-        }
         return $this->validatorChain;
     }
 
--- a/test/ArrayInputTest.php
+++ b/test/ArrayInputTest.php
@@ -45,12 +45,17 @@ final class ArrayInputTest extends TestCase
 
     protected function setUp(): void
     {
-        $this->input = new ArrayInput('foo');
+        $this->input = new ArrayInput(TestHelper::createFilterChain(), TestHelper::createValidatorChain(), 'foo');
     }
 
     protected function tearDown(): void
     {
         AbstractValidator::setDefaultTranslator();
+    }
+
+    private function createArrayInput(?string $name = null): ArrayInput
+    {
+        return new ArrayInput(TestHelper::createFilterChain(), TestHelper::createValidatorChain(), $name);
     }
 
     /**
@@ -480,7 +485,7 @@ final class ArrayInputTest extends TestCase
 
     public function testRequiredWithoutFallbackAndValueNotSetProvidesAttachedNotEmptyValidatorIsEmptyErrorMessage(): void // phpcs:ignore
     {
-        $input = new ArrayInput();
+        $input = $this->createArrayInput();
         $input->setRequired(true);
 
         $customMessage = [
@@ -716,8 +721,8 @@ final class ArrayInputTest extends TestCase
 
     public function testMergingTwoInputsModifiesTheName(): void
     {
-        $a = new ArrayInput('a');
-        $b = new ArrayInput('b');
+        $a = $this->createArrayInput('a');
+        $b = $this->createArrayInput('b');
         $a->merge($b);
 
         self::assertSame('b', $a->getName());
@@ -725,8 +730,8 @@ final class ArrayInputTest extends TestCase
 
     public function testMergingTwoInputsModifiesErrorMessage(): void
     {
-        $a = new ArrayInput('a');
-        $b = new ArrayInput('b');
+        $a = $this->createArrayInput('a');
+        $b = $this->createArrayInput('b');
         $b->setErrorMessage('Foo');
         $a->merge($b);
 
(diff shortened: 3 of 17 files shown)


src/Input.php
```diff
@@ -57,8 +58,10 @@ class Input implements
     protected $hasFallback = false;
 
     /** @param null|string $name */
-    public function __construct(protected $name = null)
-    {
+    public function __construct(
+        private readonly FilterPluginManager $filterPluginManager,
+        protected $name = null
+    ) {
```
froschdesign: Even though I believe this to be correct, it becomes a problem for the simple usage: `$email = new Input('email');`

https://docs.laminas.dev/laminas-inputfilter/v2/intro/#introduction

gsteel: The thing is, you can no longer just create a filter chain, or validator chain out of thin air - they both require plugin managers to operate, so…

```php
$input = new Input('foo');
$input->getValidatorChain()->attach(new Whatever());
```

… is not gonna work anyway.

`setValidatorChain` and `setFilterChain` are theoretically OK, however, we might need to seriously consider that you can no longer `new` an input, and you must use `Factory::create($spec)` or similar - i.e. we only support factory creation of inputs. Factory creation is better anyhow, mainly because internally, we will call `$inputManager->build(Input::class)`, so, users will need to define `Input` derivatives in DI - we cannot guarantee constructors are sound in subclasses.

I don't see this as much of a problem. I think the bigger problem will be new-ing input filters. I find it very unlikely there are many subclasses of `Input` in the wild, but I suspect there are many users subclassing `InputFilter`.

🤔

gsteel:
@ramchale & @froschdesign

`Input`s real dependencies here are the filter chain and the validator chain.

If these were constructor deps rather than the plugin managers, we would not need the plugin managers at all and there would be no hidden dependencies.

The problem of 'simple' constructors is not solved by this, but I don't believe that problem _can_ be solved very easily now that the filter/validator chains require DI up-front too.

I welcome more discussion on this, but I'm thinking that this signature is the direction we should go in:

```
public function __construct(
    protected FilterChain $filterChain,
    protected ValidatorChain $validatorChain,
) {
}
```

Paging @laminas/technical-steering-committee for comment too…

src/Factory.php
```diff
@@ -154,7 +155,12 @@ public function createInput($inputSpecification)
             ));
         }
 
-        $input = $managerInstance ?: new $class();
+        if ($class === Input::class || is_subclass_of($class, Input::class)) {
```
samsonasik: `is_a` seems can be used, see https://3v4l.org/kjoXX#v8.4.10

```suggestion
        if (is_a($class, Input::class, true)) {
```

ramchale: Thanks 😄

ramchale:
> If these were constructor deps rather than the plugin managers, we would not need the plugin managers at all and there would be no hidden dependencies.

I can see the benefit. Happy to refactor this after others have had chance to comment

Is making the new constructor parameters optional and if not set treating it like there is not a validator and/or filter until a chain is set a possibility for supporting the simple case?

(Would mean getFilterChain and getValidatorChain would now return a null if not set rather than create one)

gsteel:
> Is making the new constructor parameters optional and if not set treating it like there is not a validator and/or filter until a chain is set a possibility for supporting the simple case?
> 
> (Would mean getFilterChain and getValidatorChain would now return a null if not set rather than create one)

IMO they should be hard dependencies so `getFilterChain` can never return `null` fe.

The factory would need a refactor to generate the chains from the spec prior to `new Input()`.

I think that this may also kill off the idea of "Default" filter chains and validator chains too - These defaults are always just empty chains anyhow.

froschdesign:
> …but I don't believe that problem _can_ be solved very easily now that the filter/validator chains require DI up-front too.

But that means nothing from [the current introduction](https://docs.laminas.dev/laminas-inputfilter/v2/intro/) would work anymore and the problem is that this is often found in userland code:

```php
$inputFilter = new InputFilter();
$inputFilter->add(new Input('email'));

$input = new Input('email');

$factory = new Factory();
$inputFilter = $factory->createInputFilter([
   // …
]);

// …
```

---

I welcome the new simple and strict implementation in the code, but it will be a drastic cut if we do not provide a simple usage.

gsteel:
> I welcome the new simple and strict implementation in the code, but it will be a drastic cut if we do not provide a simple usage.

We need to find some kind of balance, I agree.

Do you think that having a hard requirement for users to use the factory like:

```php
$factory = getFactoryFromSomewhere();
$input = $factory->createInput('email');
$inputFilter = $factory->createInputFilter();
$inputFilter->add($input);
// Or
$inputFilter->add($spec, 'email');
```

… Is too big of a BC break?

The truth is that inputs and input filters have unavoidable dependencies that are _not_ trivial to orchestrate.

The current implementation means that _new_ ing these things leads to a lot of duplicated plugin manager implementations and head scratching when some factory instance cannot return a custom configured input filter for example.

i.e.

```php
(new Input())->getFilterChain()->getFilterPluginManager() !== (new Input())->getFilterChain()->getFilterPluginManager()
```

Will making `Factory` usage more ergonomic offset the inability for users to just `new` inputs and input filters?

froschdesign:
> We need to find some kind of balance, I agree.

👍🏻 

> ```
> $factory = getFactoryFromSomewhere();
> ```

I had the same example in mind, but even the first line is a problem. 😜 

> The current implementation means that new ing these things leads to a lot of duplicated plugin manager implementations and head scratching when some factory instance cannot return a custom configured input filter for example.

Or to the problem if external dependencies are required, like a database adapter. The [old DB validators](https://docs.laminas.dev/laminas-validator/v2/validators/db/) are a good example of this.

ramchale:
> even the first line is a problem

We could add a method to the FactoryFactory for this.

```php
public static function createStandaloneFactory(): Factory
{
    $serviceManager = new ServiceManager();

    $factory = new Factory(
        new FilterPluginManager($serviceManager),
        new ValidatorPluginManager($serviceManager),
        new InputFilterPluginManager($serviceManager)
    );

    $serviceManager->setService(Factory::class, $factory);

    return $factory;
}
```

Then sm v4 and v3 everything else the example becomes something like

```php
$factory = FactoryFactory::createStandaloneFactory();
$inputFilter = $factory->createInputFilter([]);

$email = $factory->createInput(['type' => Input::class, 'name' => 'email']);
$email->getValidatorChain()->attach(new EmailAddress());

$password = $factory->createInput(['type' => Input::class, 'name' => 'password']);
$password->getValidatorChain()->attach(new StringLength(['min' => 8]));

$inputFilter->add($email);
$inputFilter->add($password);
$inputFilter->setData($_POST);

if ($inputFilter->isValid()) {
    echo "The form is valid\n";
} else {
    echo "The form is not valid\n";
    foreach ($inputFilter->getInvalidInput() as $error) {
        print_r($error->getMessages());
    }
}
```

gsteel:
> We could add a method to the FactoryFactory for this.

Yes - this is what I had in mind, though I'd prefer to add a static constructor to `Factory` directly, for example:

```php
final class Factory {
    // ...

    public static function new(ContainerInterface|null $container = null): self
    {
        $container = $container ?? new ServiceManager();

        if ($container->has(self::class)) {
            return $container->get(self::class);
        }

        $factory = new Factory(
            $container->has(FilterPluginManager::class)
                ? $container->get(FilterPluginManager::class)
                : new FilterPluginManager($container),
            // etc...
        );

        if ($container instanceof ServiceManager) {
            $container->setService(self::class, $factory);
        }

        return $factory;
    }
    
    // ...
}
```

This way we could keep/make `FactoryFactory` internal and accept the maintenance burden of the new factory method in `Factory`…

WDYT @froschdesign?

Shall we just break BC on the constructor here and implement usability mitigations later, or do you think we need to keep the squishy hidden deps and lazy initialisation?

froschdesign:
@gsteel 
The factory looks very good! 👍🏻

What do you think about using Rector for migration? We already use Rector in [laminas-servicemanager-migration](https://github.com/laminas/laminas-servicemanager-migration) and it is [introduced in laminas-hydrator](https://github.com/laminas/laminas-hydrator/pull/135). Maybe some rules can be created to make it easy to rewrite old usages?

WDYT @samsonasik

samsonasik:
@froschdesign that can be doable with give diff of before vs fixture for it so I or other contributor can create custom rule for it, probably create feature request at https://github.com/laminas/laminas-servicemanager-migration/issues

ramchale:
@gsteel were you thinking the constructor params for Factory would change as well as Input?

gsteel:
@ramchale

> were you thinking the constructor params for Factory would change as well as Input?

I don't think that `Factory` requires any dependencies other than the 3 plugin managers, so, unlikely - `Factory` already depends on all 3 right?

I believe that for `Input` we should go with making `FilterChain` and `ValidatorChain` hard constructor deps - and also leave the door open for more hard constructor deps if necessary _(such as "name", "required" etc)_.

Avoid injecting plugin managers because they will attract service-locator type hidden dependency problems.

<del>We should leave out improvements to `Factory` here, and just focus on the changes to `Input`.</del> Edit… I see you've just added the static constructor already so 👍 

Sound like a plan?

ramchale:
> `Factory` already depends on all 3 right?

Yep. But as far as I can see Factory only depends on FilterPluginManager and ValidatorPluginManager to construct the FilterChain and ValidatorChain, so just double checking you don't want to move that up any further 😄 

> FilterChain and ValidatorChain hard constructor deps

This should be coverd in the most recent commit.

I think the only other question I had was whether we're removing Input->setFilterChain and setValidatorChain ?

test/TestHelper.php
```diff
@@ -13,22 +13,62 @@
...
 use LaminasTest\InputFilter\TestAsset\ValidatorStub;
+use ReflectionProperty;
 
 final class TestHelper
 {
+    private static ?ServiceManager $serviceManager                 = null;
+    private static ?FilterPluginManager $filterPluginManager       = null;
+    private static ?ValidatorPluginManager $validatorPluginManager = null;
+    private static ?Factory $factory                               = null;
+
```
gsteel: Because most of the tests will be happy with an entirely new set of dependencies for each test/required object, can we drop the static props? These will introduce complexity and subtle bugs, for example, I'd rather see:

```php
public static function validatorPlugins(ContainerInterface|null $container = null): ValidatorPluginManager
{
    $container ??= new ServiceManager();

    return new ValidatorPluginManager($container);
}
```

Because it's more explicit about how deps are handled - either you have an entirely new object graph, or you give it a specific container when you need cohesion between collaborators.

Later on, when we have more integration tests, we can add a helper here to create a container such as:

```php
public static function getContainer(array $config = []): ContainerInterface
{
    $defaults = (new ConfigProvider())->__invoke();
    $config = array_merge_recursive($defaults, $config);
    $config['dependencies']['services']['config'] = $config;
    
    return new ServiceManager($config['dependencies']);
}
```

And then fetch whatever you need for a given test from that container.

Static props give me the shivers 👻

test/ArrayInputTest.php
```diff
@@ -45,7 +45,7 @@ final class ArrayInputTest extends TestCase
 
     protected function setUp(): void
     {
-        $this->input = new ArrayInput('foo');
+        $this->input = new ArrayInput(TestHelper::createFilterChain(), TestHelper::createValidatorChain(), 'foo');
```
gsteel, line 48: I've got no objection to what you've done here and in all the other tests with `new Input($deps…)`, but, as we refactor further, if constructor args change more, this will get tiresome to maintain, just looking at it gives me the feeling that a local method such as `$this->arrayInputWithName('foo')`, per `TestCase` might be a good idea now or in the future 👍

src/Factory.php
```diff
@@ -154,7 +184,19 @@ public function createInput($inputSpecification)
...
-        $input = $managerInstance ?: new $class();
+        if (is_a($class, Input::class, true)) {
+            $filterChain = new FilterChain();
+            /** @psalm-suppress DeprecatedMethod removal will be done in Service Manager 4 upgrade */
+            $filterChain->setPluginManager($this->filterPluginManager);
+
+            $validatorChain = new ValidatorChain();
+            $validatorChain->setPluginManager($this->validatorPluginManager);
+
+            /** @psalm-suppress UnsafeInstantiation */
```
gsteel, line 195: At some point, we should introduce:

```php
$input = match($class) {
    Input::class,
    ArrayInput::class => new $class($deps…),
    default => $inputFilterPluginManager->build($class, $opts),
};
```

Or, we might decide to introduce a factory or abstract factory and just call `$inputFilterPluginManager->build($class, $opts)`.

The `UnsafeInstantiation` here really mustn't be ignored…

src/Input.php
```diff
@@ -56,9 +50,11 @@ class Input implements
     /** @var bool */
     protected $hasFallback = false;
 
-    /** @param null|string $name */
-    public function __construct(protected $name = null)
-    {
+    public function __construct(
+        protected FilterChain $filterChain,
+        protected ValidatorChain $validatorChain,
+        protected ?string $name = null
```
gsteel, line 56: This might actually be `array-key` for `$name` - Tests are passing, but IIRC, once an input is retrieved from a collection, it's name can end up being `(int) 0`.

We can worry about that later though.

gsteel (changes requested):
This looks good to me @ramchale - Thank you!

I'd like to see changes to static props in the test helper, but otherwise just comments and thoughts from me.

I think we should retain `setFilterChain` and `setValidatorChain` for now with a view to potentially dropping them in future patches… I can just about see a use case for

```php
$input = $factory->createInput('foo');
$input->setFilterChain($chainBuiltFromConfigSomehow);
```

Also, we may end up with immutable chains, so being able to replace the chain entirely would be necessary in that scenario.

test/ArrayInputTest.php
ramchale: I've added a factory method to any test classes that do this more than a few times

gsteel (approved):
Thanks for all your work here @ramchale 👍

Could you do a force push here before merge though? Doesn't look like GH triggered the CI run…

PR #149, opened 2025-08-07
Service manager v4

### LINKED ISSUE #121: Implement service manager 4
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | TBD

#### Summary

Upgrade the dependency to service manager 3 OR 4, or only v4, in case there are some BC breaks 

related to https://github.com/dotkernel/api/issues/265

ramchale:
This still needs a lot of tidy up, but seems fairly close to a working version with service-manager v4.

Raised as a draft to check if there's anything else obvious I'm missing, and seek some help fixing the remaining tests.

(The block of commented out code in Factory being an obvious area that needs a look)


gsteel:
Just playing around locally with this patch - and validator ^3.6 will defo be needed so that we can `$validators->build(ValidatorChain::class, $spec)`.

I think we might need to do some more cleanup before we can figure out what's tripping up those tests… 🤔

ramchale:
Thanks for the work on resolve the underlying changes @gsteel

I've just done a quick rebase and fixed a few obvious bits. Busy day ahead, so I'll get back to this when I can. Happy for anyone else to make progress with it in the meantime if they'd like to

gsteel:
Cheers @ramchale 👍

test/InputFilterAbstractServiceFactoryTest.php
```diff
@@ -54,78 +55,82 @@ protected function setUp(): void
...
+
+        $validator = $this->createMock(ValidatorInterface::class);
+        $validatorPluginManager->configure(['services' => ['foo_validator' => $validator]]);
+
+        // Temporary assertions to check configuration
+        self::assertSame($filter, $filterPluginManager->get('foo_filter'));
+        self::assertSame($validator, $validatorPluginManager->get('foo_validator'));
+        // End
+
+        $inputFilter = $this->factory->__invoke($services, 'filter');
```
ramchale, line 138: This fails with "ServiceNotFoundException: Unable to resolve service "foo_validator" to a factory"

Debugging thorugh, it looks like the filter also can't be resolved but does not throw an Exception.
The build methods of the filterPluginManager and validatorPluginManager don't seem like they try to use the 'services', so I'm not sure this can pass unless we want to make further changes to those.

Worth double checking though, as I'm much less familiar with the internals of the plugin managers than some other parts of the code

gsteel: > The build methods of the filterPluginManager and validatorPluginManager don't seem like they try to use the 'services',

Yes! This is what had me a bit stumped… `build()` will never consult `services`, only `get()` will.

I've added fixes to both `filter` and `validator` which should help us move forward:

- https://github.com/laminas/laminas-filter/pull/261
- https://github.com/laminas/laminas-validator/pull/435

ramchale: Thanks @gsteel
Looks like it gets over that hurdle 😄

gsteel: Tests should pass once https://github.com/laminas/laminas-validator/pull/437 is merged.

There are a number of skipped tests that focus on runtime validation of validator/filter spec arrays - runtime validation of these is gone, so, we need to figure out what to do there…

ramchale: @gsteel Been a while as I've got some life stuff going on, but yep the tests are looking a lot better. Thanks.
Is the intent for things like FactoryTest->testCreateInputWithInvalidFilterSpecType that static analysis is relied on to pick up these, so we can remove the test? At least in the cases where we've needed to add `/** @psalm-suppress InvalidArgument */`

gsteel: Hey @ramchale 👋

I've been bitten recently removing defensive runtime checks in favour of static analysis. I think you have to look at that sort of stuff on a case-by-case basis.

For things like array spec - you have to know the spec up-front by reading docs, so, if you get that wrong as a user, I think it's OK to have `TypeError` rather than defensive runtime checks, but also, it's easy to make mistakes with array specs and clear exception messages really help there.

I hope the life stuff is _good_ life stuff 👍

ramchale: We can add some type checking in input-filter as in the [latest commit ](https://github.com/laminas/laminas-inputfilter/pull/149/commits/18b1fc7925aba0cc39a8287ea242bd73a4162a2b), but a couple of the skipped tests feel like the type validation more naturally sit in laminas-filter and laminas-validator, closer to where the original type error the test looks for originated. I can put in PRs to those repositories add checks for the presence of the required array keys with more helpful error messages if that works for you?

gsteel:
@ramchale We've got green CI 🎉

Is there anything else to do in this patch or shall we review and get it merged?

src/Factory.php
```diff
@@ -88,39 +90,37 @@ private function canCreateInputType(string $class): bool
...
-            $filterChain->setPluginManager($this->filterPluginManager);
-        } else {
-            $filterChain = new FilterChain();
-            $filterChain->setPluginManager($this->filterPluginManager);
-            $this->populateFilters($filterChain, $filters);
+        $filters    = $spec['filters'] ?? [];
+        $validators = $spec['validators'] ?? [];
+
+        if (! is_array($filters) && ! is_callable($filters) && ! $filters instanceof FilterInterface) {
+            throw new TypeError("filters must be an array, callable, or FilterInterface. Received: "
```
gsteel: This should not be a `TypeError`, it should be `InvalidArgumentException` in this NS

src/Factory.php
```diff
@@ -88,39 +90,37 @@ private function canCreateInputType(string $class): bool
...
-     * @psalm-suppress DeprecatedMethod removal will be done in Service Manager 4 upgrade
      * @param InputSpecification $spec
      * @return array{
-     *     filterChain: FilterChain,
-     *     validatorChain: ValidatorChain,
+     *     filterChain: FilterChainInterface,
+     *     validatorChain: ValidatorChainInterface,
      * }
      */
     private function buildChainsFromSpecification(array $spec): array
```
gsteel, line 100: We should probably add all the `@throws` here so it's easier to amend the public API to include `@throws` where necessary

test/FactoryTest.php
```diff
@@ -103,23 +97,20 @@ public function testCreateInputWithTypeAsAnInvalidClassInstanceThrowException():
         ]);
     }
 
-    public function testCreateInputWithFiltersAsAnInvalidTypeThrowException(): void
+    public function testCreateInputWithInvalidFilterSpecType(): void
     {
         $factory = $this->createDefaultFactory();
-
         $this->expectException(TypeError::class);
```
gsteel: This test should be checking for a specific exception type in our NS

ramchale:
Awesome work @gsteel !

Aside from squashing the commits, no. I think anything else can be a seperate PR

test/FactoryTest.php
```diff
@@ -287,7 +273,10 @@ public function testFactoryWillCreateInputWithSuggestedFilters(): void
                     break;
                 case 2:
                     self::assertInstanceOf(Filter\StringToLower::class, $filter);
-                    self::assertEquals('iso-8859-1', $filter->getEncoding());
+                    self::assertEquals(
+                        'iso-8859-1',
+                        (new ReflectionObject($filter))->getProperty('encoding')->getValue($filter)
+                    );
```
gsteel: This test needs completely re-working to get rid of reflection introduced because of dropped upstream getters.

For example:

1) Create a filter chain with a predicatable output
2) Assert that executing the filter yields the desired output

We can assert the size of the chain with `assertCount(2, $chain)` because I think it's `Countable`

test/FactoryTest.php
```diff
@@ -299,16 +288,16 @@ public function testFactoryWillCreateInputWithSuggestedFilters(): void
     public function testFactoryWillCreateInputWithSuggestedValidators(): void
     {
         $factory = $this->createDefaultFactory();
-        $digits  = new Validator\Digits();
+        $digits  = new Digits();
```
gsteel, line 295: Same thing with the equivalent validator chain test, make the chain then assert the chain fails or passes validation for given inputs based on validators added and options.

test/InputFilterAbstractServiceFactoryTest.php
```diff
@@ -54,102 +59,108 @@ protected function setUp(): void
...
-        self::assertFalse($this->factory->canCreate($this->services, 'filter'));
-    }
+        $services = new ServiceManager();
 
-    public function testCannotCreateServiceIfConfigServiceDoesNotHaveInputFiltersConfiguration(): void
-    {
-        $this->services->setService('config', []);
+        if ($config !== null) {
+            $services->setService('config', $config);
+        }
```
gsteel: Instead of `setService`, we should inject into the constructor with `new ServiceMananager(['services' => ['config' => $config]]);`

test/InputFilterPluginManagerCompatibilityTest.php
```diff
@@ -38,41 +41,33 @@ protected static function getPluginManager(array $config = []): InputFilterPlugi
 
     public function testShareByDefaultAndSharedByDefault(): void
```
gsteel:
```suggestion
    public function testInputFiltersAreNotSharedByDefault(): void
```

gsteel (changes requested):
Spotted a few nitpicks

test/InputFilterAbstractServiceFactoryTest.php
gsteel: Or conditionally inject it when non-null, i.e `$options = $config === null ? [] : [...]`

src/Factory.php
ramchale: Happy to change it, but currently testCreateInputWithInvalidFilterSpecType is testing for TypeError so it's a slight change to the expected behaviour

ramchale:
> Spotted a few nitpicks

Thanks. I'll pick most of those up tomorrow

src/Factory.php
gsteel: I think it was me that introduced the `TypeError` changes in this patch and previously it was something else.

ramchale:
@gsteel think I've covered off all the feedback. Let me know if there's anything else you can see or if it's good to sqash

gsteel (approved):
Thanks @ramchale LGTM!

PR #150, opened 2025-08-07
Replace InputFilterPluginManagerFactory deprecated interface

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | yes
| QA            | no

### Description

Replace the deprecated Laminas\ServiceManager\ServiceLocatorInterface with Laminas\ServiceManager\Factory\FactoryInterface

### DIFF
--- a/src/InputFilterPluginManagerFactory.php
+++ b/src/InputFilterPluginManagerFactory.php
@@ -4,13 +4,9 @@ declare(strict_types=1);
 
 namespace Laminas\InputFilter;
 
-use ArrayAccess;
-use Laminas\ServiceManager\Config;
-use Laminas\ServiceManager\FactoryInterface;
-use Laminas\ServiceManager\ServiceLocatorInterface;
+use Laminas\ServiceManager\Factory\FactoryInterface;
 use Laminas\ServiceManager\ServiceManager;
 use Psr\Container\ContainerInterface;
-use Psr\Container\ContainerInterface as PsrContainer;
 
 use function assert;
 use function is_array;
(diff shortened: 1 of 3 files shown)


src/InputFilterPluginManagerFactory.php
```diff
@@ -19,27 +16,15 @@
  * @link ServiceManager
  *
  * @psalm-import-type ServiceManagerConfiguration from ServiceManager
- *
- * @final
```
gsteel, line 23: Lets make this internal too

src/InputFilterPluginManagerFactory.php
```diff
@@ -19,27 +16,15 @@
...
-    {
+    /** @param string|null $requestedName */
+    public function __invoke(
+        ContainerInterface $container,
+        $requestedName = null,
+        ?array $options = null
+    ): InputFilterPluginManager {
         $pluginManager = new InputFilterPluginManager($container, $options ?? []);
 
         // If this is in a laminas-mvc application, the ServiceListener will inject
```
gsteel: Can you please drop anything related to 'ServiceListener' here, along with tests.
MVC projects will never get to install this!

gsteel: Also, further down (can't comment on unchanged lines), drop the check for `ArrayAccess` and just assert $config is an array IMO.

gsteel (changes requested):

src/InputFilterPluginManagerFactory.php
ramchale: > drop the check for `ArrayAccess`

Makes sense 👍

gsteel (approved):
Thanks @ramchale 👍

PR #151, opened 2025-08-07
Refactor `Factory::createInput()`

gsteel:
- Factory now refuses to `new` custom `Input` types. We don't know what an extending class constructor looks like because they are not subject to LSP rules.
- `createInput` can no longer return an `InputFilter` - this brings significant improvements to type inference for users. A `create` method is used to conditionally return an `Input|InputFilter`
- "Default" filter chains, and "Default" validator chains are gone, along with the setters and getters.

Also, pretty big changes to the `InputFilter` Tests… Some of the data providers were hellish to read and follow and uneccessarily complicated.

The plan here is that after the upgrade, we should be closer to green CI - and hopefully failing tests in #149 will be fixed, or more easily fixable 🤞

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -46,7 +46,6 @@
     </TooManyArguments>
     <UnusedPsalmSuppress>
       <code><![CDATA[RedundantConditionGivenDocblockType]]></code>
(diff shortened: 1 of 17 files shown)


ramchale (approved):
Cool. Looks more straight forward to me

froschdesign (approved):

PR #154, opened 2025-10-02
Specify types for isValid $context and return value

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | yes
| QA            | no

### Description

Improve the type safety and compatibility with laminas-validator v3

### DIFF
--- a/src/InputInterface.php
+++ b/src/InputInterface.php
@@ -105,10 +105,7 @@ interface InputInterface
      */
     public function getValue();
 
-    /**
-     * @return bool
-     */
-    public function isValid();
+    public function isValid(): bool;
 
     /**
      * @return array<array-key, string>
--- a/src/ArrayInput.php
+++ b/src/ArrayInput.php
@@ -48,7 +48,7 @@ class ArrayInput extends Input
     }
 
     /** @inheritDoc */
-    public function isValid($context = null)
+    public function isValid(?array $context = null): bool
     {
         $hasValue    = $this->hasValue();
         $required    = $this->isRequired();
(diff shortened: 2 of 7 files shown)


src/InputInterface.php
```diff
@@ -105,10 +105,7 @@ public function getValidatorChain();
      */
     public function getValue();
 
-    /**
-     * @return bool
-     */
-    public function isValid();
+    public function isValid(): bool;
```
gsteel, line 108: Why would we not add the context parameter to the interface when all implementations use it?

I think it can be argued that all `Input`s could be considered internal _(Kind of)_, so given that `InputFilter` is responsible for calling `Input::isValid()` we can pretty much guarantee that `$context` is always available, so I'd be tempted to interface the method with `isValid(array<string, mixed> $context): bool`.

WDYT?

ramchale: > Why would we not add the context parameter to the interface when all implementations use it?

Thanks for taking a look.

It is the sensible thing to do, but it opens a can of worms with psalm errors so not sure if we want to address that in one PR.  Very short version is that if we're going to change that, we should probably also change `InputFilterInterface` to match, unless we want to surpress some errors for now.

That mostly looks okay with the exception of the array merge of the context with the `$data` member variable in `BaseInputFilter` (https://github.com/laminas/laminas-inputfilter/blob/2.34.x/src/BaseInputFilter.php#L250). 
`$data` specifices `array-key` instead of `string`, which I think might be valid but conflicts with the downstream signature for `ValidatorChain->isValid` that prompted this (https://github.com/laminas/laminas-validator/blob/3.9.x/src/ValidatorChain.php#L215)

PR #155, opened 2025-10-13
Drop support for PHP 8.1, add support for PHP 8.5

gsteel:


Xerkus:
So we have a bit of a weird case here.

Tests data provider uses invalid `null` for `$name` in 

https://github.com/laminas/laminas-inputfilter/blob/63e2e94cb8e71a923bcc0b7e72c6fbe5aa3718bd/src/BaseInputFilter.php#L119-L171

In php 8.5 it causes deprecation. What should we do to fix it?

gsteel (commented):
LGTM @Xerkus - Happy to merge and release?

Xerkus:
Give me a sec. I want to update docs workflow and get it merged up asap on release

PR #156, opened 2025-10-14
Add a guard against invalid input names in input filter

Xerkus:
Input filter does not properly handle invalid input filter names. Problematic behavior is exhibited when input name is null or empty string and input instance has null or empty string for the name.

```php
$input1 = new Input();
$input2 = new Input();
// ... some code
$inputFilter->add($input2); // set as ''
$inputFilter->add($input2); // merged into '' - input 1
$inputFilter->has(null); // true
$inputFilter->get(null); // same as get('') and returns input 1
```

Both inputs have null as an input name that effectively is treated as `''`. 
Quite unexpected `$input` instance was modified with values from `$anotherInput`. Yikes.

https://github.com/laminas/laminas-inputfilter/blob/63e2e94cb8e71a923bcc0b7e72c6fbe5aa3718bd/src/BaseInputFilter.php#L103-L112

Ironically, tests are setup with invalid dataset in data provider and it came to light only thanks to PHP 8.5 deprecation. See #155 

Let's practice Source Driven Test Development :+1:. Tests are unreadable so I start with introduction of the guards in the source to make tests fail.

### DIFF
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -20,6 +20,7 @@ use function func_get_args;
 use function get_debug_type;
 use function is_array;
 use function is_int;
+use function is_string;
 use function sprintf;
 
 /**
@@ -98,6 +99,29 @@ class BaseInputFilter implements
 
         if ($input instanceof InputInterface && ($name === null || $name === '' || is_int($name))) {
             $name = $input->getName();
+
+            /** @psalm-suppress DocblockTypeContradiction Input conflicts with InputInterface docblock and allows null. */
+            if ($name === null || $name === '') {
+                throw new Exception\InvalidArgumentException(sprintf(
+                    '%s: input instance must have a valid name or input name must be provided as a parameter',
+                    __METHOD__,
+                ));
+            }
+        }
+
+        if (! is_string($name) && ! is_int($name)) {
+            throw new Exception\InvalidArgumentException(sprintf(
+                '%s: input name expected to be string or int, %s given',
+                __METHOD__,
+                get_debug_type($name)
+            ));
+        }
+
+        if ($name === '') {
+            throw new Exception\InvalidArgumentException(sprintf(
+                '%s: input name can not be an empty string',
+                __METHOD__,
+            ));
         }
 
         if (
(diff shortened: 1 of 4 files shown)


src/BaseInputFilter.php
```diff
@@ -98,6 +99,28 @@ public function add($input, $name = null)
...
+                __METHOD__,
+                get_debug_type($name)
+            ));
+        }
+
+        if ($name === '') {
+            throw new Exception\InvalidArgumentException(sprintf(
+                '%s: input name can not be an empty string',
+                __METHOD__,
+            ));
```
gsteel, line 124: Can we extract this into a method such as

```
/** @psalm-assert non-empty-string|int $name */
private function assertValidInputName(string|int|null $name, string $callingMethod): void {}
```

Xerkus: I don't think it is worth it. We will just drop this in 3.0.0

Xerkus:
@gsteel I have some questions for you https://github.com/laminas/laminas-inputfilter/pull/91

https://github.com/laminas/laminas-inputfilter/blob/63e2e94cb8e71a923bcc0b7e72c6fbe5aa3718bd/test/InputFilterTest.php#L112-L123

gsteel:
@Xerkus Pretty sure I added that test to verify existing behaviour, just for my sanity. It has nothing to do with the patch, so I assume I was trying to ensure I didn't break anything.

I think the test can be removed as it just documents broken existing behaviour

Xerkus:
tests are harder to understand than the code they test...

src/BaseInputFilter.php
```diff
@@ -167,6 +209,14 @@ public function get($name)
...
+        /** @psalm-suppress DocblockTypeContradiction  */
+        if (! is_string($name) && ! is_int($name)) {
+            throw new Exception\InvalidArgumentException(sprintf(
+                '%s: input name expected to be string or int, %s given',
+                __METHOD__,
+                get_debug_type($name),
+            ));
+        }
         return array_key_exists($name, $this->inputs);
     }
```
Xerkus, line 221: We can return false here instead of an exception. In case someone was relying on the bad behavior.

gsteel: I think the exception is fine. Same for `remove()`.

gsteel (approved):
LGTM. Thanks @Xerkus

PR #160, opened 2025-11-17
Remove support for the MVC Module Manager

gsteel:
MVC users _could_ still use and install this library in the future by making use of the config provider to configure the service manager


Ocramius:
Conflicted about this: are modules now confined to the MVC concept? :thinking:

gsteel:
I've been removing them elsewhere - can @froschdesign or @Xerkus confirm with regard to the status of `module-manager`?

froschdesign:
Modules based on laminas-modulemanager are only used in laminas-mvc. Modular applications can also be created in Mezzio, but laminas-config-aggregator ([config providers](https://docs.laminas.dev/laminas-config-aggregator/config-providers/)) is used here. And this was also planned for version 4 of laminas-mvc: [Drop module manager and add support for config aggregator](https://github.com/laminas/laminas-mvc/pull/56).
Furthermore, all new versions of components that support laminas-servicemanager version 4 cannot be used with laminas-mvc, because it still requires version 3 of laminas-servicemanager.
The laminas-modulemanager itself is in security-only maintenance mode [since last August](https://github.com/laminas/technical-steering-committee/blob/main/meetings/minutes/2025-08-04-TSC-Minutes.md#mark-mvc-related-libraries-as-security-only-in-line-with-laminas-mvc).

froschdesign (approved):

gsteel:
Thanks @froschdesign for the explanation

PR #161, opened 2025-11-17
`ExceptionInterface` should inherit from `Throwable`

gsteel:


Ocramius (approved):
LGTM, but IMO could land in 2.x too

gsteel:
I'll send in a duplicate patch for 2.x along with a load of deprecations 👍

PR #162, opened 2025-11-17
Minor BC Breaking QA

gsteel:
- Final exceptions
- Match factory signatures to supers
- Add readonly keyword where possible


Ocramius (approved):
LGTM

PR #163, opened 2025-11-17
Require `psr/http-message` as a direct dependency

gsteel:
We use these symbols in `/src` so they should be listed as a hard dependency


Ocramius (approved):
LGTM

PR #167, opened 2025-11-18
Simplifies and consolidates documentation structure; removes laminas-mvc integration example

froschdesign:
First clean-up. Removes also the `module_class` entry from MkDocs configuration for installation page.


docs/book/v3/application-integration/usage-in-a-mezzio-application.md
```diff
@@ -1,16 +1,13 @@
...
 
-The following example shows _one_ potential use case of laminas-inputfilter within
-a mezzio based application. The example uses a module, config provider
-configuration, laminas-servicemanager as dependency injection container, the
-laminas-inputfilter plugin manager and a request handler.
+The following example shows _one_ potential use case of laminas-inputfilter within a Mezzio-based application.
+The example uses a module, config provider configuration, laminas-servicemanager as a dependency injection container, the laminas-inputfilter plugin manager and a request handler.
 
-Before starting, make sure laminas-inputfilter is [installed and configured](../installation.md).
+Before starting, make sure the lamina-inputfilter is [installed and configured](../installation.md).
```
gsteel:
```suggestion
Before starting, make sure the laminas-inputfilter is [installed and configured](../installation.md).
```

docs/book/v3/application-integration/usage-in-a-mezzio-application.md
```diff
@@ -64,27 +60,21 @@ use Psr\Http\Message\ResponseInterface;
...
-    
     public function __construct(
-        InputFilterInterface $inputFilter,
-        TemplateRendererInterface $renderer
-    ) {
-        $this->inputFilter = $inputFilter;
-        $this->renderer    = $renderer;
-    }
+        private InputFilterPluginManager $inputFilterPluginManager
+        private TemplateRendererInterface $renderer
```
gsteel, line 67: For an example, Injecting the plugin manager is OK I guess, but I would normally inject the desired input filter, otherwise we're in "hidden dependencies" territory. I don't know if it's worth doing it that way for this example or not though.

docs/book/v3/application-integration/usage-in-a-mezzio-application.md
```diff
@@ -96,83 +86,45 @@ class ListHandler implements RequestHandlerInterface
...
-```
-
-> ### Instantiating the InputFilter
+> INFO: **Instantiating the Input Filter**
 >
-> The `InputFilterPluginManager` is used instead of directly instantiating the
-> input filter to ensure to get the filter and validator plugin managers
-> injected. This allows usage of any filters and validators registered with
-> their respective plugin managers.
+> The input filter plugin manager (`Laminas\InputFilter\InputFilterPluginManager`) is used instead of directly instantiating the input filter to ensure to get the filter and validator plugin managers injected.
```
gsteel:
```suggestion
> The input filter plugin manager (`Laminas\InputFilter\InputFilterPluginManager`) is used instead of directly instantiating the input filter to ensure that the filter and validator plugin managers are injected correctly.
```

gsteel (changes requested):

docs/book/v3/application-integration/usage-in-a-mezzio-application.md
froschdesign: > …but I would normally inject the desired input filter, otherwise we're in "hidden dependencies" territory.

Correct, but he `Laminas\ServiceManager\AbstractFactory\ReflectionBasedAbstractFactory` does not support resolving forms, validators, filters, input filters, etc., as these cannot be retrieved directly from the service container. And creating a factory is not user-friendly or even effective for advertising. 😉 
Therefore I prefer [reinfi/zf-dependency-injection](https://github.com/reinfi/zf-dependency-injection?tab=readme-ov-file#what-can-be-autowired) but this creates an extra dependency which we do not want in these examples. This would be better suited to a cookbook recipe.
Therefore, this compromise has been reached.

gsteel (approved):
Thanks @froschdesign :)

docs/book/v3/application-integration/usage-in-a-mezzio-application.md
froschdesign: Please do not get me wrong, because documentation is not an advertising brochure, but we are measured by how simple or complicated the examples are. This even affects the entire component.

gsteel: @froschdesign 

I totally agree - the lengths you have to go to inject _just_ the input filter, for the purposes of a "simple" example are extreme.

I was thinking yesterday about how we should ship reflection based factories that work with any/all plugin managers in our own ecosystem and came to the conclusion that it would have to be a standalone library with soft dependencies.

Much as I hate soft deps, it may be worth it for a library that ships one class, i.e.

`UniversalReflectionBasedAbstractFactory` that will resolve:

- Validators
- Filters
- InputFilters
- Inputs
- ViewHelpers
- Regular, top-level deps
- Other plugin-manager things…

The only _clean_ way to do this would be to make a pluggable factory via `config` with `ReflectionBasedDependencyResolverInterface` or some such, so `view` ships this class that gets a constructor arg from reflection and reports whether it can resolve it or not, but, life is too short, so, soft-deps would be like

```php
if (class_exists(InputFilterInterface::class) && is_a($arg, InputFilterInterface::class, true)) {
   $plugins = $container->get(InputFilterPluginManager::class);
   $constructorDep = $plugins->get($argName);
}
```

WDYT?

froschdesign: > …we should ship reflection based factories that work with any/all plugin managers in our own ecosystem…

That would be very nice! 👍🏻 
I had planned to use autowiring entirely in a new getting started tutorial, [registered as an abstract factory](https://docs.laminas.dev/laminas-servicemanager/v4/configuring-the-service-manager/#abstract-factories) for everything. And at the end, there is a chapter on performance which explains how to [generate factories using CLI commands](https://docs.laminas.dev/laminas-servicemanager/v4/cli-commands/generate-ahead-of-time-factories/).

---

The resolver concept of reinfi/zf-dependency-injection is a good solution:

- https://github.com/reinfi/zf-dependency-injection/blob/master/src/Service/AutoWiring/Resolver/PluginManagerResolver.php
- https://github.com/reinfi/zf-dependency-injection/blob/master/src/Injection/AutoWiringPluginManager.php
- https://github.com/reinfi/zf-dependency-injection/tree/master/src/Service/AutoWiring/Resolver 

> ##### What can be autowired?
> 
> Every service registered within the service manager can be autowired.
> Plugins within the plugin manager can also be autowired. If you need to register another mapping, you can add the following:
> 
> ```php
> PluginManagerResolver::addMapping('MyInterfaceClass', 'MyPluginManager');
> ```
> 
> If your service needs the container as a dependency, this can also be autowired.
> 
> ##### Add another resolver
> 
> If you like to add another resolver, you can add one through the configuration.
> 
> ```php
> 'reinfi.dependencyInjection' => [
>     'autowire_resolver' => [
>         AnotherResolver::class,
>     ],
> ]
> ```
> It needs to implement the ResolverInterface.

PR #174, opened 2025-12-29
Introduce `ErrorMessages` Type

gsteel:
The nested structure of error messages is difficult to manage and somewhat squishy…

Validators always return `array<string, string>`, but input filter error message arrays can be nested indefinitely - these types of recursive structures are a nightmare for static analysis and can easily be mutated unintentionally.

An object is easier to reason about, easier to process recursively in a type-safe way and easier to refactor and identify.

Additionally, we can get an accurate count of actual error messages instead of counting empty arrays - i.e. `count($inputFilter->getMessages())` will be zero, even if it contains an empty set of errors for each valid input.

Array access has been retained, so users will still be able to access the same error messages as the same indexes, regardless of how deeply nested the result is.

This type may also be useful in future for translation purposes

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -17,14 +17,9 @@
   <file src="src/BaseInputFilter.php">
     <InvalidReturnStatement>
       <code><![CDATA[$values]]></code>
-      <code><![CDATA[array_map(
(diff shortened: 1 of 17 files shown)


froschdesign (approved):
That makes sense to me!

PR #178, opened 2026-01-10
Feature Removal: Runtime mutation of filter chain and validator chain

gsteel:
Now that the filter chain and validator chain are constructor dependencies, setters are no longer necessary, so the following methods have been removed from all implementations of `InputInterface`

- `setFilterChain()`
- `setValidatorChain()`

### DIFF
--- a/src/Input.php
+++ b/src/Input.php
@@ -74,12 +74,6 @@ class Input implements
         return $this;
     }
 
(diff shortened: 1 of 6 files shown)


froschdesign (approved):
Everything fine! 👍🏼

PR #179, opened 2026-01-16
Make the `Input` name a constructor dependency

gsteel:
This patch makes it impossible to create an input with an empty name.

Name is now a constructor dependency, and it is validated to be `non-empty-string|int`


froschdesign (approved):
LGTM 👍🏻

PR #180, opened 2026-01-17
Split `InputInterface` and add missing methods

gsteel:
`InputInterface` is too big, so all mutating methods are moved into a new interface `MutableInputInterface` - this is mostly the setters and the `merge` method.

`EmptyContextInterface` is deleted and its methods re-allocated between `InputInterface` and `MutableInputInterface`

Additionally, the methods on `Input` relating to "Fallback Values" are added to these interfaces where appropriate.

The reason for this re-organisation is that it is _very_ likely that most users type hint and consume the concrete class, whereas at the point of use, `InputInterface` should only be required.

In future `MutableInputInterface` can be deprecated and removed as we move toward immutability.

The concrete input gains an `$options` parameter that is used to seed the various props that `Factory` used to do at runtime after construction paving the way to remove the setters entirely from the concrete classes.


froschdesign (approved):
LGTM 👍🏻

PR #189, opened 2026-06-10
Deprecate the 'old' api for removal in 4.0

gsteel:


src/InputInterface.php
```diff
@@ -13,28 +13,53 @@ interface InputInterface
     /** @internal */
     public const EMPTY_FAILURE_VALIDATION_KEY = '__inputEmptyValueFailure';
 
+    /**
+     * @deprecated Since 3.0. Setting the value to be validated is deprecated as part of the old stateful API.
+     *             Use the {@link validate()} method instead.
```
froschdesign, line 18: Is it sufficient if the new method is only mentioned in the interface?

gsteel: Good call @froschdesign - I looked into this with PHPStorm, and the extended deprecation message needs to be duplicated to the implementations

froschdesign (approved):
LGTM 👍🏼

